PR #6, opened 2019-12-20
Dev liam

liamhuber:


pyiron_contrib/protocol/utils/comparers.py
```diff
@@ -161,5 +160,13 @@ def equals(self, b):
         return all(conditions)
 
 
+class ListComparer(Comparer):
 
+    type = list
+
+    def equals(self, b):
+        assert isinstance(b, list)
```
dgehringer: Here instead of the `assert` statement, I think it would be better to return `False`, since`assert` will raise an `AssertionError` if we have type mismatch. And the programm crashes

PR #11, opened 2020-02-07
Images

liamhuber:
A basic prototype for handling image data inside pyiron. No unittests yet, but an educational demo notebook is in the new 'demos' directory.


samwaseda:
I'll try to do the review tomorrow....

liamhuber:
No worries! I chose you because you made the fatal mistake of showing interest earlier this week ;) That and because Ning hasn't accepted the pyiron github invite yet.

Since you might want to show it to other colleagues I thought we could kill two birds with one stone by having you familiarize yourself with it at the same time as we find any problem code I snuck in.

PR #13, opened 2020-07-10
Remove image demo

liamhuber:
There is an Aachen-specific demo on the Aachen cluster. The image code is only preliminary anyhow, so right now there is no need to maintain a second demo here.


jan-janssen:
In the interest of https://github.com/orgs/pyiron/teams/experimental I would like to collect the pyiron examples for experimental applications somewhere here on GitHub. We could even think about creating one repository with all examples to make it easier for future presentations.

PR #21, opened 2020-10-04
Elastic constants

liamhuber:
New job types for calculating elastic constants given chemistry. Uses `SQSJob` by @jan-janssen and @dgehringer, and `ElasticMatrixJob` (currently private in pyiron_mpie) by @yury-lysogorskiy.

This is not super polished, but I want it merged to the master soon anyhow since @muh-hassani will make use of it in the software taskforce meeting to show how we can couple pyiron and DAMASK all in one notebook, and I don't want him to have to dig around different branches.

The structure is a core job for using one concentration, and a list one that wraps this and runs over a list of possible concentrations. The key output is the actual structures, their chemical compositions and the elastic matrices for those compositions. In principle there's more data buried in the HDF5, but those are thing only things with nice bindings so far.

The last big issue is that I inherit only from `GenericJob`, not from `ParallelMaster` or `FlexibleMaster`. Maybe there is a clever way to work with these, but I ran into trouble because new `ElasticMatrixJob` instances needed to be created from a reference job, e.g. `lammps_ref_instance.create_job(...)`, and I couldn't work out a way with either of those `Master` workflows to adjust the necessary parameters. So it's not simply that I'm ignorant of their existence.

Known issues:
- Only running with modal run mode.
- Helper functions break on loaded jobs -- this is because `InputList` is wrapping lists and dicts as `InputList(dict)` etc. in HDF. I'll make an issue of it over in pyiron_base when I understand it a bit better.
- The SQS job seems to be ignoring the number of structures. The input is registering correctly in HDF, so I think the problem might be with `SQSJob`. Will investigate further and raise an issue as needed.
- Needs docstrings. I'll add these ASAP, but need sleep now.
- Most output and analysis is real python objects, not inspecting the HDF5. This can obviously be slow, but I'm not worried about optimizing for speed there yet.


jan-janssen:
I am confused, why are you using `GenericJob` rather than `ParallelMaster` as base class? From my perspective the workflow is very similar to https://github.com/pyiron/pyiron/blob/master/notebooks/pipeline.ipynb

jan-janssen (changes requested):
As this pull request reimplements the logic of the `Parallelmaster` I am strongly against merging it and prefer to discuss it in the Friday meeting. Just recently I created an example to implement your own `Parallelmaster` and I currently see no reason why this should not be implemented based on a `Parallelmaster `https://github.com/pyiron/pyiron/blob/master/notebooks/gamma_surface_parallel_master.ipynb .

jan-janssen:
Ok, maybe I should have read your comment more than your code. To address the point that elastic matrix is created from the lammps job that is the same as create_job from the project https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/job/generic.py#L1012 and then setting the reference job. The other component is the StructureListMaster which is used in the pipeline to calculate the phonons for each volume of the Murnaghan https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/master/structure.py it accepts a list of structures and the reference job can be a ParallelMaster which inside contains another reference job.

liamhuber:
I agree that there should be `ParallelMaster` inheritance here, all I can say is that I tried for several hours each for the `SQSElasticConstants` early on and later `SQSElasticConstantsList` and never managed to get it to run without failure and using the desired input. I'm not saying it can't be done, just that with hours of effort I didn't manage.

I'm very open to revising this and rebasing it onto `ParallelMaster` going forward, but there simply is not time for that now; this is only the first step of a coupled pyiron+DAMASK workflow. I would like to have this coupling presented at the software taskforce meeting on Friday, and I need to give Muhammad and Vitesh time to integrate it with the second part. (This is still just an internal demo, before showing such coupling to the institute we'll still have time to modify the code.) I would rather merge something (to contrib! not to core pyiron) that is still in need of improvement than to force them to play around with a sub-branch of the repo.

Unfortunately, because of the SFB1394 general assembly this week, I don't expect to be at the pyiron meeting this week. I'd be happy to discuss it in more detail the following week, at which point I may also have had an opportunity to rebase the inheritance.

jan-janssen:
To me there is a fundamental difference between what is on the master branch and what is used for presentations. Presentations can either be based on additional modules/ python files which are distributed with the presentation of the code could even be included in the notebook to make it self consistent. But the master branch is something which is shared with collaborators and we should not reimplement something which is already in the main code, unless there is a strong reason to do so. My fear is that once we merge it here we are going to have more and more parallel developments, simply saying it was done here as well. Therefore the presentation on Friday is no reason to merge it from my perspective.

jan-janssen:
I was thinking about this again and noticed the same parallele master could be used also for the calculation of the free energy for phase diagrams https://github.com/jan-janssen/NiCr-pyiron-example  we just need a master for the SQS jobs the rest should already be possible with the structure master and the phonopy job or the elastic matrix accordingly.

liamhuber:
Yeah, that's a good idea! I think in general there's probably a number of routines we can wrap in SQS+ListMaster.

I managed to get it working for this case (see branch head) but won't pull until I get some better output processing written too.

PR #22, opened 2020-10-09
Setup & Installation

jan-janssen:
In preparation to migrate parts of pyiron_mpie to pyiron_contrib we should implement the same setup scripts we use for pyiron.


jan-janssen:
Currently I get the following errors: 
```
======================================================================
ERROR: test_atoms (protocol.utils.test_comparers.TestComparer)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/pyiron/pyiron_contrib/tests/protocol/utils/test_comparers.py", line 81, in test_atoms
    self.assertTrue(Comparer(a) == a)
  File "/home/travis/build/pyiron/pyiron_contrib/pyiron_contrib/protocol/utils/comparers.py", line 118, in __eq__
    return self._equals(other)
  File "/home/travis/build/pyiron/pyiron_contrib/pyiron_contrib/protocol/utils/comparers.py", line 89, in _equals
    return self.default(b) if comparer is None else comparer(self.object).equals(b)
  File "/home/travis/build/pyiron/pyiron_contrib/pyiron_contrib/protocol/utils/comparers.py", line 178, in equals
    return all(conditions)
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
======================================================================
ERROR: test_lists (protocol.utils.test_comparers.TestComparer)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/pyiron/pyiron_contrib/tests/protocol/utils/test_comparers.py", line 95, in test_lists
    self.assertTrue(Comparer(a) == a)
  File "/home/travis/build/pyiron/pyiron_contrib/pyiron_contrib/protocol/utils/comparers.py", line 118, in __eq__
    return self._equals(other)
  File "/home/travis/build/pyiron/pyiron_contrib/pyiron_contrib/protocol/utils/comparers.py", line 89, in _equals
    return self.default(b) if comparer is None else comparer(self.object).equals(b)
  File "/home/travis/build/pyiron/pyiron_contrib/pyiron_contrib/protocol/utils/comparers.py", line 194, in equals
    all([Comparer(val) == last_val for val, last_val in zip(self.object, b)])
  File "/home/travis/build/pyiron/pyiron_contrib/pyiron_contrib/protocol/utils/comparers.py", line 194, in <listcomp>
    all([Comparer(val) == last_val for val, last_val in zip(self.object, b)])
  File "/home/travis/build/pyiron/pyiron_contrib/pyiron_contrib/protocol/utils/comparers.py", line 118, in __eq__
    return self._equals(other)
  File "/home/travis/build/pyiron/pyiron_contrib/pyiron_contrib/protocol/utils/comparers.py", line 89, in _equals
    return self.default(b) if comparer is None else comparer(self.object).equals(b)
  File "/home/travis/build/pyiron/pyiron_contrib/pyiron_contrib/protocol/utils/comparers.py", line 178, in equals
    return all(conditions)
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
----------------------------------------------------------------------
```

liamhuber:
> Currently I get the following errors:
> ...
> ```
> ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
> ...
> ```

Yep, me too. I'll look into it.

liamhuber:
The test should now be [fixed](https://github.com/pyiron/pyiron_contrib/pull/24). At the end of the day it was just the update from `structure.cell` to `structure.cell.atoms`.

liamhuber (commented):
Looks sharp! Two more general questions

1) Here we include the conda version of pyiron, but stuff I do in pyiron_contrib is typically running on the bleeding edge of the main pyiron repo -- if we are relying on bleeding edge behaviour, will this cause our CI tools to fail? Forgive me if we're already pointing to the pyiron/master rather than the conda version somewhere, I searched for this but might have overlooked/not recognized it.

2) I guess that pyiron_base makes it into the dependency tree via pyiron, but in general there's nothing about contrib that implies it follows from pyiron rather than pyiron_contrib. Should we additionally explicitly include pyiron_base in the requirements?

setup.py
```diff
@@ -0,0 +1,43 @@
+"""
+Setuptools based setup module
+"""
+from setuptools import setup, find_packages
+
+
+setup(
+    name='pyiron_contrib',
+    version='0.0.1',
+    description='Repository for the pyiron plugins developed at the Max Planck Institute für Eisenforschung.',
```
liamhuber, line 10: Rather 'Repository for user-generated plugins to the pyiron IDE developed by the Max Planck Institute für Eisenforschung'.

I guess the most widely used plugin (in terms of geography: Austria+Canada) is Dominik's QM/MM protocol, and he's not an MPIE memeber.

jan-janssen: I copied the file from `pyiron_mpie`, now I phrased it `Repository for user-generated plugins to the pyiron IDE`

jan-janssen:
> Looks sharp! Two more general questions
> 
> 1. Here we include the conda version of pyiron, but stuff I do in pyiron_contrib is typically running on the bleeding edge of the main pyiron repo -- if we are relying on bleeding edge behaviour, will this cause our CI tools to fail? Forgive me if we're already pointing to the pyiron/master rather than the conda version somewhere, I searched for this but might have overlooked/not recognized it.

I now install pyiron as well as pyiron_base directly from `pip --pre` which equals the master branch. 

> 2. I guess that pyiron_base makes it into the dependency tree via pyiron, but in general there's nothing about contrib that implies it follows from pyiron rather than pyiron_contrib. Should we additionally explicitly include pyiron_base in the requirements?

I plan to move parts of pyiron_mpie to pyiron_contrib and those definitely depend on pyiron rather than pyiron_base and I guess the same is true for parts of the protocols. So for now I guess we include both in den dependencies, otherwise we would need to split the repository in something like pyiron_contrib and pyiron_base_contrib but I do not think that is necessary now.

liamhuber:
> So for now I guess we include both in den dependencies, otherwise we would need to split the repository in something like pyiron_contrib and pyiron_base_contrib but I do not think that is necessary now.

Perfect! Yes, I was definitely aiming for "both" rather than "instead". I hope one day pyiron_contrib will also have dependencies on pyiron_experiment, pyiron_continuum, etc :) 

Whenever the checks pass I'm happy with it.

liamhuber (approved):

PR #25, opened 2020-10-09
Migrate structure jobs to pyiron_contrib

jan-janssen:


liamhuber (commented):
These come with zero documentation -- what is the difference/advantage of these over the existing SQSJob? 

For contrib I certainly don't think we need tests and complete docstring coverage, but I think it would be wise for public-facing classes to at least have a class-level docstring.

jan-janssen:
ATAT includes high order terms like cluster expansion and should give more reliable results, still the issue is that it does not stop but just continuous running and has to be stopped at some point. So from my perspective we can also remove the ATAT job completely. 
In contrast randspg can generate structures for a given space group based on the idea that high symmetry structures are lower in energy. The idea is to use it to identify stable phases even for interatomic potentials which might have ghost phases: 
https://www.sciencedirect.com/science/article/pii/S0010465516303848

liamhuber:
Perfect, let's kill atatsqs then.

Cool, I'm happy to keep randspg in this case. Could you add a super short docstring to it? I think even something like the sentence and paper ref you just gave me is already sufficient. Basically I think the user should be able to get *some* idea of what the class does without digging into the names/functionality of its methods or straight up googling it -- this is especially important for classes with abbreviations in the name. At the same time, the bar for getting stuff into contrib should stay reasonably low. The level of description you gave is a good balance of these, IMO.

jan-janssen:
Maybe we should add an example notebook like https://github.com/pyiron/pyiron-meeting-mpie/blob/master/handbook/1-structures/1-4-structure-generators/randspg-lammps-library.ipynb

PR #28, opened 2020-10-16
SQS Elastic constants

liamhuber:
As part of coupling with DAMASK, I made a new job type which strings together SQS, minimization, and elastic constants (stored it the private repo pyiron_mpie). You can either run a single instance, or a ParallelMaster that handles multiple of these jobs with different compositions.

In it's current state these jobs run OK, save, and load alright -- i.e. they're operational. But, they're definitely still not pretty. There's many `TODO` items throughout the source code showing where I think the development should go, so please ctrl+F and scan through these to see if I'm already aware of the issue.

Things I'd like help with before merging:
- HDF management. 
  - This is the big one. Right now the HDF storage is very flat and ugly in the project. It's trickier than the usual setup, because my job has multiple reference jobs instead of just one. 
  - The first thing I'd like help with is resolution of [Issue 51](https://github.com/pyiron/pyiron_base/issues/51) discussion with @jan-janssen over in pyiron_base, where `copy_to` does not function properly. 
  - Once that's resolved I can take another crack at cleaning this up, but I'd welcome other suggestions for HDF/child management.
- @pmrv's very excellent InputList needs some beefing up as we move it towards being `IOList`
  - The main thing is the [stale issue 42](https://github.com/pyiron/pyiron_base/issues/42) over at pyiron_base, that InputList wraps complex datatypes also as InputLists when saving them, then doesn't know how to convert them back to a non-InputList on loading. 
  - It would also be nice to get InputList able to handle things like lists of Structures
  - Also to try to call `to/from_hdf` on it's children in case they have that already. It may be helpful to look at Dominik's generic hdf code, which I extracted [here](https://github.com/pyiron/pyiron_contrib/blob/abstract_graph/pyiron_contrib/utils/hdf.py) in a different contrib branch.

Other things on my radar and I will get added (eventually) before merging:
- Sometimes the SQS job only produces one structure instead of the requested number. This seems to happen when `mole_fractions` is given, but not when we rely on the input structure composition. I'm not sure if it's a problem with how I'm using it or with SQSJob, but I'll look into it.
- Sometimes the output is not loading properly, e.g. `residual_pressures` in the demo notebook.
- Adding some unit tests. I really liked Sam's point (somewhere?) that at least we should have a unit test that instantiates the job!


jan-janssen:
The calculation of SQS structures for multiple concentrations is a general concept, which for example is also used for the calculation of temperature concentration phase diagrams:
https://github.com/jan-janssen/NiCr-pyiron-example

Therefore I suggest to implement a separate ParallelMaster for this, like it Is done in:
https://github.com/pyiron/pyiron-meeting/blob/master/2020-10-16/jan-janssen/sqsmaster.ipynb

liamhuber:
> Therefore I suggest to implement a separate ParallelMaster for this, like it Is done in:
> https://github.com/pyiron/pyiron-meeting/blob/master/2020-10-16/jan-janssen/sqsmaster.ipynb

Yeah, sounds good. Should be a matter of breaking out part of the existing setup and moving it over to sqs.py.

raynol-dsouza (approved):

muh-hassani:
[In this line,](https://github.com/pyiron/pyiron_contrib/blob/569e059a28c1e68c58a68c65c4f8d1486c65bcc7/pyiron_contrib/atomistic/atomistics/master/sqselastic.py#L11) the elastic module from pyiron_mpie is imported. Since pyiron_mpie is private, it would be better to import only elastic.py locally. I have done the same thing in [here](https://github.com/muh-hassani/pyiron_workflow_examples/blob/main/LAMMPS-DAMASK-workflow/pyiron_contrib/atomistic/atomistics/master/sqselastic.py)

liamhuber:
> [In this line,](https://github.com/pyiron/pyiron_contrib/blob/569e059a28c1e68c58a68c65c4f8d1486c65bcc7/pyiron_contrib/atomistic/atomistics/master/sqselastic.py#L11) the elastic module from pyiron_mpie is imported. Since pyiron_mpie is private, it would be better to import only elastic.py locally. I have done the same thing in [here](https://github.com/muh-hassani/pyiron_workflow_examples/blob/main/LAMMPS-DAMASK-workflow/pyiron_contrib/atomistic/atomistics/master/sqselastic.py)

I'll convert it over to Sam's new elastic job before merging, which will also resolve this issue

PR #29, opened 2020-10-27
Remove hdf assignment already handled by the super() call

niklassiemer:
After the discussion in [pyiron_base](https://github.com/pyiron/pyiron_base/pull/50#issue-502077335) I also got rid of the additional hdf if clauses in the image job.

### DIFF
--- a/pyiron_contrib/image/job.py
+++ b/pyiron_contrib/image/job.py
@@ -184,14 +184,12 @@ class ImageJob(GenericJob):
             group_name (str): HDF5 subgroup name - optional
         """
         super(ImageJob, self).to_hdf(hdf=hdf, group_name=group_name)
-        if hdf is None:
-            hdf = self.project_hdf5
-        self.input.to_hdf(hdf=hdf, group_name=None)
-        self.output.to_hdf(hdf=hdf, group_name=None)
-        with hdf.open("images") as hdf5_server:
+        self.input.to_hdf(hdf=self._hdf5, group_name=None)
+        self.output.to_hdf(hdf=self._hdf5, group_name=None)
+        with self._hdf5.open("images") as hdf5_server:
             for n, image in enumerate(self.images):
                 image.to_hdf(hdf=hdf5_server, group_name="img{}".format(n))
-        hdf["n_images"] = n + 1
+        self._hdf5["n_images"] = n + 1
 
     def from_hdf(self, hdf=None, group_name=None):
         """
@@ -202,8 +200,6 @@ class ImageJob(GenericJob):
             group_name (str): HDF5 subgroup name - optional
         """
         super(ImageJob, self).from_hdf(hdf=hdf, group_name=group_name)
-        if hdf is None:
-            hdf = self.project_hdf5
         self.input.from_hdf(hdf=self._hdf5, group_name=None)
         self.output.from_hdf(hdf=self._hdf5, group_name=None)
         with self._hdf5.open("images") as hdf5_server:


liamhuber:
What I really get scared by is that the `hdf` argument is now *not used*. Either your code is doing what it's supposed to and there is a heinous problem upstream that we have meaningless variables in our signature, or your code should use `_hdf5` only as a backup when `hdf=None`.

I suspect your code is perfect, which is actually unfortunate in this case ;)

jan-janssen:
At @liamhuber it is used in `super(ImageJob, self).to_hdf(hdf=hdf, group_name=group_name)`

niklassiemer:
As @jan-janssen pointed out, `hdf` is used in the `super` call. There `self._hdf` is set to the provided `hdf` if not None. Therefore, we were checking this at each layer while it was already taken care of.

liamhuber:
Aha, so the goal is to eventually get rid of this `hdf` argument everywhere? 

If we pass in a non-None `hdf`, why don't we also want to use it in this method? Is this just a slightly dangerous naming convention, where we have an argument `hdf` but then don't use it for the `hdf` argument of sub-methods called?

jan-janssen:
The argument for the current solution is that the code developer only needs to take care of `self._hdf` or `self.project_hdf` and does not need to take care of the check if `hdf` is set or not. A more clean solution would be to have static methods which just take `hdf` and do not use `self._hdf` or `self.project_hdf` and then basically use a decorator which handles the switch in case `hdf` is not set.

niklassiemer:
If we pass in a non-None `hdf` we will use exactly this hdf, since `self._hdf=hdf` in this case due to the `super` call. Of course we could retain the `if hdf==None: hdf = self._hdf`, however after the `super` call one could also just write `hdf=self._hdf`

liamhuber:
> If we pass in a non-None `hdf` we will use exactly this hdf, since `self._hdf=hdf` in this case due to the `super` call. 

Ahhahaha, aye, there's the rub. I accept this behaviour, but I definitely don't like it.

```
class Foo:
    def meth(self, arg):
        self._arg = arg

class Bar(Foo):
    def meth(self, arg):
        super().meth(arg)
        return self._arg + 2
```

^^ This is certainly legal, but for developers who are working on `Bar` or child classes thereof it's super cruel.

liamhuber:
Given that this is the reality of the situation though, I now very clearly see the benefit of not re-checking the `if-None` case and am 100% behind this PR.

liamhuber (approved):
Given that this is the reality of the situation though, I now very clearly see the benefit of not re-checking the if-None case and am 100% behind this PR.

niklassiemer:
> Ahhahaha, aye, there's the rub. I accept this behaviour, but I definitely don't like it.
> 
> ```python
> class Foo:
>     def meth(self, arg):
>         self._arg = arg
> 
> class Bar(Foo):
>     def meth(self, arg):
>         super().meth(arg)
>         return self._arg + 2
> ```
> 
> ^^ This is certainly legal, but for developers who are working on `Bar` or child classes thereof it's super cruel.

Do I understand correctly, that you say it is cruel since the variable `self._arg` of `class Bar` appears out of nowhere?

liamhuber:
> Do I understand correctly, that you say it is cruel since the variable `self._arg` of `class Bar` appears out of nowhere?

Yes, that's all I mean. To me it is just a violation of "explicit is better than implicit". Here we pass in an argument, but then never use it in the code we're presently writing, but rather only pass it to a super which silently updates some other attribute and *that* is then the piece of information we're *actually* using in our new code. It's a legal architecture choice but...

confusing

liamhuber:
And I should be clear: I don't *like* it, but I don't have any better ideas at the moment either so, like 0% judgement here. Just something we should keep in the back of our minds as a yellow flag.

PR #35, opened 2020-12-06
adding FEniCS job

muh-hassani:
adding FEniCS job to pyiron.


niklassiemer:
Could you add your new job to the `JOB_CLASS_DICT` in __init__.py of contrib such that
```
from pyiron import Project
import pyiron_contrib # Adds pyiron_contrib JobTypes
```
Adds the the fenics job type to the usual list of available jobtypes. 
I would also like to see the fenics module bound to some property of the Job such that the import section is only
```
from pyiron_base import Project
import pyiron_contrib
import numpy as np
import matplotlib.pyplot as plt
```
and then you call `p = job.fenics.Expression(...)`

niklassiemer:
Looks way nicer! However, I wonder if we shouldn't after all bind fenics to a property of our fenics-job, since I assume that we lack a lot of functionality in general. So if someone used to fenics wants to do some other stuff he or she has to import fenics after all. 
Therefore, I would add 
```
@property 
def fenics(self):
    return FEN
```

pyiron_contrib/continuum/fenics.py
```diff
@@ -195,8 +195,13 @@ def run_static(self):
         Solve a PDE based on 'LHS=RHS' using u and v as trial and test function respectively. Here, u is the desired
         unknown and RHS is the known part.
         """
+        self.status.running = True
         self.u = FEN.Function(self.V)
         FEN.solve(self.LHS == self.RHS, self.u, self.BC)
+        self.collect_output()
```
niklassiemer, line 201: The `collect_output` is triggered by `run()` if `self.status.collect`, thus
```
self.status.collect = True
self.run()
```

liamhuber: Yes, this is true. I did it the other way because it felt more explicit, but you're absolutely right that the way you suggest is more consistent with the code base. I fixed it.

However, this brings me back again to the point that we need to revisit our pyiron_base architecture here -- these canonical methods should all be wrapped with something that automatically updates the status so that we don't duplicate, e.g. `self.status.collect = True; self.run()` in (almost) every subclass of `GenericJob`!!!

muh-hassani:
@liamhuber thanks for your latest update to the notebook. But there is still a dependency on pyiron in pyiron atomistic from `pyiron_contirib.__init__.py` :
https://github.com/pyiron/pyiron_contrib/blob/3db457259483b75ecf7b2ffc1e1715d46a8a4b39/pyiron_contrib/__init__.py#L4

pyiron_contrib/__init__.py
```diff
@@ -1,7 +1,11 @@
 __version__ = "0.1"
 __all__ = []
 
-from pyiron import Project
+try:
+        from pyiron import Project
+except:
+        print("Warning: pyiron module not found;\nBe carefule not to use \"from pyiron_contrib import Projects\"")
```
jan-janssen, line 7: Please do not use print but instead warnings.warn() in addition we should import the project from pyiron_base.

muh-hassani:
Done!
https://github.com/pyiron/pyiron_contrib/blob/f51c4a6991afa1511a36bfc2e6ae5b0ad861ed86/pyiron_contrib/__init__.py#L9

pyiron_contrib/__init__.py
muh-hassani: Done!
https://github.com/pyiron/pyiron_contrib/blob/f51c4a6991afa1511a36bfc2e6ae5b0ad861ed86/pyiron_contrib/__init__.py#L9

liamhuber (approved):
I can't figure out how to remove myself as a reviewer, so I'll just switch to approving ;) 

But seriously, thanks for adding the changes we discussed at the meeting @muh-hassani! The TODO item in the docstring should be good enough, and now every instantiation throws a warning -- perfect. With these disclaimers in place I think it's ready for review now.

.ci_support/test_environment.yml
```diff
@@ -0,0 +1,15 @@
+channels:
+  - conda-forge
```
jan-janssen, line 2: Please use the same naming used in the user pyiron modules to be consistent - https://github.com/pyiron/pyiron/blob/master/.ci_support/environment-notebooks.yml

.github/workflows/integrationtest.yml
```diff
@@ -0,0 +1,38 @@
+
```
jan-janssen, line 1: e.g. https://github.com/pyiron/pyiron/blob/master/.github/workflows/notebooks.yml

jan-janssen: I am fine with having the integration tests additionally - but then we should also add the notebooks test just like we have it in the other packages.

.ci_support/test_environment.yml
muh-hassani: Do you mean the name of the file should be changed to `environment-test.yml`?

.github/workflows/integrationtest.yml
muh-hassani: I will add it.

.ci_support/test_environment.yml
jan-janssen: I would name it `environment-notebooks.yml`

liamhuber:
@niklassiemer and @jan-janssen, I'm happy to merge this now and I don't think we need re-review from you, but I also don't want to merge it before you're ready in case you wanted to take another look.

jan-janssen (approved):
Looks good to me

PR #36, opened 2020-12-09
Introducing a file browser GUI

niklassiemer:
This introduces a file browser GUI, which supports:
 - Browsing of the local file system with preview of some supported file types (images, txt, svg)
 - Browsing a S3 based object storage with preview of the metadata corresponding to the respective objects
 - Inspecting of H5 files like directories with preview of the stored data ( `FileHDFio()[key]` )
Each of the features may be (de)activated:
The H5 files may be treated as plain files or completely be hidden.
Switching between the S3 and the local file system may be restricted as well as switching the current directory (only for the S3 system).
This file browser is exposed at the (pyiron_contrib) project level in the mode which allows only the access to the local file system and browsing of H5 files like directories:
```
from pyiron_contrib.project.project import Project
pr=Project("TestFileBrowser")
pr.open_file_browser()
```
Be aware that this implementation of `Project` is inherited from the pyiron implementation, however, adds a new h5 file (`self.base_name + "_projectdata"`) to allow for storing of some metadata.


niklassiemer:
As discussed today, I will refactor the file browser to be tightly bound to the project.

liamhuber:
@niklassiemer I'm not digging deep right now, but I just want to say I appreciate the clarity of your commit messages, it's easy to skim over them and get a quick sense of what's being done 🙂

niklassiemer:
Since I started to push this to base, I close this PR.

PR #39, opened 2020-12-12
Fix bug in from_hdf

raynol-dsouza:
Fixed a small bug in `from_hdf` that prevented the protocols from running on the queue.


liamhuber (commented):
So getting/keeping this running on the queue is absolutely critical and if that change does this then I'm in favour, but I definitely don't understand what we're doing here. (With full recognition that I'm pretty sure I'm the one who wrote this mess of `to/from_hdf` for `Protocol` to begin with!)

`Protocol` inherits from both `CompoundVertex` and `GenericJob`. We can (as of python3?) simply call `super().method(*args, *kwargs)`, and [my understanding](https://stackoverflow.com/questions/3277367/how-does-pythons-super-work-with-multiple-inheritance) is that for multiple inheritance this tries to find `method` in the left-most parent and runs that.

Further, I think a method of `Protocol` calling `super(Protocol, self).*` should be equivalent to `super().*`. So I'm very confused that your change has an effect.

I have one minor concern and one major concern:

# Minor

Right above this in `Protocol.to_hdf`, we have a call `super(CompoundVertex, self).to_hdf` similar to what you now do here -- I like this symmetry. But in `to_hdf` we're manually to-HDFing the `graph` variable, yet here `from_hdf` we don't go out of our way to touch the graph. I think that since `CompoundVertex.to_hdf` actually does the graph business, the line `self.graph.to_hdf(hdf=hdf, group_name="graph")` in `Protocol.to_hdf` is redundant. While we're synchronizing `to_hdf` and `from_hdf` appearance here, it would be nice to get that resolved.

# Major

So, like, we *never* call `GenericJob.to/from_hdf`? It's been a looong time since I've look at this code, but that seems kind of confusing and inefficient at best (like, we must re-implement some of the behaviour `GenericJob` stores for these methods), or flat out wrong and problematic at worst. Do you have any insight into this and how/why we're getting away with it?

raynol-dsouza:
@liamhuber I tried a few things before changing `super(Protocol, self).from_hdf(hdf=hdf, group_name=group_name)` to `super(CompoundVertex, self).from_hdf(hdf=hdf, group_name=group_name)` (including re-applying the ABC inheritance), but nothing worked, except for this. I did not dig in too deep as to why this works just yet.

I also think that the way we store output needs a revamp, especially now that more and more pyiron jobs are using Marvin's DataList (InputList?). I believe Dominic implemented a self-consistent way of storing and accessing all of the input/output which bypasses `GenericJob` altogether. But again, this is one area of the Protocols that I have not spent a lot of time on, especially because it has been working so far without a hitch..

pmrv:
Hint: you can use the `__mro__` attribute defined on each class to lookup in which order `super()` works its way along the inheritance chain.  If you give a class argument to it, `super()` skips to that part of the chain.  It might just be that some of the classes in there forget to call `super()` or in the wrong order.  

Then again, I'm totally green when it comes to pyiron_contrib, so it might also be not helpful.

raynol-dsouza:
`Protocol.__mro__` gives the following order of the inheritance chain:

`(pyiron_contrib.protocol.generic.Protocol,
 pyiron_contrib.protocol.generic.CompoundVertex,
 pyiron_contrib.protocol.generic.Vertex,
 pyiron_contrib.protocol.utils.misc.LoggerMixin,
 abc.ABC,
 pyiron_base.job.generic.GenericJob,
 pyiron_base.job.core.JobCore,
 pyiron_base.generic.template.PyironObject,
 object)`

while using `type(Protocol)` gives `abc.ABCMeta`.

@liamhuber do you see any anomaly here? Is the order not as intended?

@pmrv thanks for the assist!

liamhuber:
> @liamhuber do you see any anomaly here? Is the order not as intended?

This is indeed what I expected. I'm still a little surprised the following does not work:

```python
Protocol(CompoundVertex, GenericJob):

    ...

        hdf = hdf or self.project_hdf5

    def from_hdf(self, hdf=None, group_name=None):
        hdf = hdf or self.project_hdf5
        super().from_hdf(hdf=hdf, group_name=group_name)
```

Could you please give that a try? If this fails I will simply accept my ignorance and we can carry on with the PR as it stands at time of writing.

pmrv:
If this works, try switching 
```python
class Protocol(CompoundVertex, GenericJob):
  ...
class Protocol(GenericJob, CompoundVertex):
```
because setting `self._project_hdf5` should be done in GenericJob.

raynol-dsouza:
@liamhuber I tried what you suggested, and it didn't work. It continues to work if I remove `self.graph.to_hdf(hdf=hdf, group_name="graph")`, provided `super(CompoundVertex, self).to_hdf(hdf=hdf, group_name=group_name)` is still present.

@pmrv switching `class Protocol(CompoundVertex, GenericJob):` to `class Protocol(GenericJob, CompoundVertex):` throws a new error, claiming certain attributes (like `input`, which `CompoundVertex `inherits from its parent) are not present.

pmrv:
> @pmrv switching `class Protocol(CompoundVertex, GenericJob):` to `class Protocol(GenericJob, CompoundVertex):` throws a new error, claiming certain attributes (like `input`, which `CompoundVertex `inherits from its parent) are not present.

Mh, I'm out of my depth then, honestly.

liamhuber:
> @liamhuber I tried what you suggested, and it didn't work. It continues to work if I remove `self.graph.to_hdf(hdf=hdf, group_name="graph")`, provided `super(CompoundVertex, self).to_hdf(hdf=hdf, group_name=group_name)` is still present.

Ok cool, I'm glad the explicit graph saving is indeed not needed. Putting `CompoundVertex` in the `super` call is perfectly acceptable to me if that's what it takes to get the desired behaviour.

> @pmrv switching `class Protocol(CompoundVertex, GenericJob):` to `class Protocol(GenericJob, CompoundVertex):` throws a new error, claiming certain attributes (like `input`, which `CompoundVertex `inherits from its parent) are not present.

Yeah, that makes perfect sense if it is calling the `GenericJob.to_hdf` instead. So at least here it works as expected.

> Mh, I'm out of my depth then, honestly.

Frankly so am I, this is very strange behaviour. I'm sure it's a smell that something deeper should be fixed, but this close to Christmas let's just be pragmatic and merge what gives the behaviour we want... 🎄

raynol-dsouza:
@liamhuber Shall I merge this pull request?

liamhuber:
Yep, go for it! It's a little unfortunate we don't know *why* it needs to be the way it is, but the behaviour is what we're looking for 👍

raynol-dsouza:
Well, next time if anything hdf-related breaks down, we can maybe take a look at it again!

PR #42, opened 2020-12-22
Try fenics without restrictions

jan-janssen:


jan-janssen:
@liamhuber The fenics conda package seems to be fixed now, so it is no longer necessary to define a specific version number or even build number.

PR #45, opened 2021-01-15
Update protocols to run with pyiron_atomistics

raynol-dsouza:


pyiron_contrib/protocol/compound/thermodynamic_integration.py
```diff
@@ -2163,4 +2136,4 @@ def get_tild_integrands(self, integrands='harm_to_inter'):
 
 
 class ProtocolVacancyFormation(Protocol, VacancyFormation):
-    pass
+    pass
```
liamhuber, line 2139: Nit: IDE nags that PEP8 wants a new line at the end of the file

liamhuber (commented):
No objections from my end. I like the renaming of zero_k energy to eq as well. Tests should be passing before merge, but otherwise go for it. I'll go skim the tests now to see if I have any insight there.

liamhuber:
Ok, the test failure is just because the testing environment can't import `pyscal`. Please cross reference with `pyiron_atomistics` to see the syntax for how to include this dependency in `setup.py` and `.ci_support/environment.yml`, then re-run the tests. At this point I bet the only thing failing will be coverage (which we can let slide for this PR unless you see something nice to cover!) and you can merge.

Aha, please also edit these files to use `pyiron_atomistics` instead of `pyiron`.

Finally, `pyiron_atomistic` in turn relies on `pyiron_base`, so the code will all run fine, but from the perspective of transparency I also lean towards explicitly include `pyiron_base` in the dependencies. What do you think. @pmrv @niklassiemer what's your opinion here?

niklassiemer:
I think we should include base explicitly, since we rely directly on `pyiron_base` in some of our import statements.

niklassiemer:
I just looked it up, in the conda pyiron version we request in our `.ci_support/environment.yml` there was no pyscal dependency, yet. I will update the environment.

pmrv:
> Finally, `pyiron_atomistic` in turn relies on `pyiron_base`, so the code will all run fine, but from the perspective of transparency I also lean towards explicitly include `pyiron_base` in the dependencies. What do you think. @pmrv @niklassiemer what's your opinion here?

Without looking into all the imports here, I think transitive dependencies should only be listed if you directly import them.

raynol-dsouza:
> Ok, the test failure is just because the testing environment can't import pyscal. Please cross reference with pyiron_atomistics to see the syntax for how to include this dependency in setup.py and .ci_support/environment.yml, then re-run the tests. At this point I bet the only thing failing will be coverage (which we can let slide for this PR unless you see something nice to cover!) and you can merge.

I guess @niklassiemer did this in another pull request. Shall I merge this one then @liamhuber ?

liamhuber:
Perfect. Since some modules in contrib explicitly import from base it sounds like we're all on the same page about including it in the dependencies. I merged Niklas's PR so just pull it in here 🙂

pmrv:
Anything still open that blocks us from merging this?

raynol-dsouza:
@pmrv I do not think so. Should be good to merge.

PR #46, opened 2021-01-20
Update environment.yml

niklassiemer:


.ci_support/environment.yml
```diff
@@ -1,13 +1,14 @@
...
-  - pyiron =0.3.6
-  - scipy =1.5.2
-  - seaborn =0.11.0
-  - scikit-image =0.17.2
\ No newline at end of file
+  - matplotlib >=3.2.1
+  - numpy >=1.19.4
+  - pyiron_base >=0.1.44
+  - pyiron_atomistics >=0.2.2
+  - scipy >=1.6.0
```
jan-janssen, line 12: I guess all these dependencies are already included in `pyiron_atomistics` https://github.com/conda-forge/pyiron_atomistics-feedstock/blob/master/recipe/meta.yaml#L25 so defining them here is redundant. The same applies for `ase`.

niklassiemer: That is true. However, what is best practice here in genera? If I need dependencies of another package - should I include them also in my list or not?

PR #72, opened 2021-02-08
Some more AtomicRex fixes 

jan-janssen:
But it is not yet working.

### DIFF
--- a/pyiron_contrib/atomistic/atomicrex/atomicrex_job.py
+++ b/pyiron_contrib/atomistic/atomicrex/atomicrex_job.py
@@ -11,7 +11,7 @@ from pyiron_base import GenericJob, Settings, PyironFactory, Executable
 
 from pyiron_contrib.atomistic.atomicrex.general_input import GeneralARInput, AlgorithmFactory
 from pyiron_contrib.atomistic.atomicrex.structure_list import ARStructureList
-from pyiron_contrib.atomistic.atomicrex.potential_factory import ARPotFactory
+from pyiron_contrib.atomistic.atomicrex.potential_factory import ARPotFactory, AbstractPotential
 from pyiron_contrib.atomistic.atomicrex.output import Output
 from pyiron_contrib.atomistic.atomicrex.function_factory import FunctionFactory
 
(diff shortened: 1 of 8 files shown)


pyiron_contrib/atomistic/atomicrex/atomicrex_job.py
```diff
@@ -44,23 +44,21 @@ def __init__(self, project, job_name):
...
+
 
     def to_hdf(self, hdf=None, group_name=None):
-        if hdf == None:
-            hdf = self.project_hdf5
         super().to_hdf(hdf=hdf, group_name=group_name)
-        self.input.to_hdf(hdf=hdf)
-        self.potential.to_hdf(hdf=hdf)
-        self.structures.to_hdf(hdf=hdf)
+        self.input.to_hdf(hdf=self._hdf5)
```
pmrv, line 51: This also needs to pass the `group_name`.  When I look at the resulting HDF5 files it seems that the type information gets overwritten by subsequent calls from members.

jan-janssen: The `group_name` of the job type is already handled by the `super()`, so we could add the group_names of the specific classes, but that would be redundant.

pmrv: We might have to specify explicit groups the members to write themselves into, but that should be taken care of by the table_name of the input lists.  I'm looking at this right now.

pyiron_contrib/atomistic/atomicrex/potential_factory.py
```diff
@@ -8,50 +8,56 @@
...
+
     @staticmethod
     def lennard_jones_potential(sigma, epsilon, cutoff, species={"a": "*", "b": "*"}, identifier="LJ"):
         return LJPotential(sigma, epsilon, cutoff, species=species, identifier=identifier)
 
-class LJPotential():
-    def __init__(self, sigma, epsilon, cutoff, species, identifier):
+
+class AbstractPotential(InputList):
+    def __init__(table_name="potential"):
```
pmrv: Missing `self` argument.

jan-janssen: Yes, I fixed this in https://github.com/pyiron/pyiron_contrib/pull/72/commits/c196a7f73be9b13d8c5d1b08313f66aff74be0f5

pyiron_contrib/atomistic/atomicrex/structure_list.py
```diff
@@ -150,7 +150,8 @@ def from_hdf(self, hdf=None, group_name="arstructure"):
             structure = Atoms()
             structure.from_hdf(hdf_s_lst, group_name="structure")
             self.structure = structure
-            self.fit_properties = hdf_s_lst["fit_properties"]
+            self.fit_properties = ARFitPropertyList()
+            self.fit_properties.from_hdf(hdf=hdf_s_lst, group_name="fit_properties")
```
pmrv, line 154: This `from_hdf` fails because `ARFitPropertyList` doesn't accept a container anymore in its `__init__`, which is something the `InputList` code (reasonably) assumes all its subclasses can.  Will look into a work-around after dinner.

jan-janssen: That is fixed in https://github.com/pyiron/pyiron_contrib/pull/72/commits/ed54ffe188d3dd24470bf4c1c806da5d06692f9c

jan-janssen:
@Leimeroth  The reloading is fixed in this branch and it is ready to merge. The example notebook is again available on https://github.com/jan-janssen/pyiron-atomicrex/tree/atomicrex_jan

pyiron_contrib/atomistic/atomicrex/atomicrex_job.py
jan-janssen: Exactly, if the `table_name` of the input list is set, then it is not necessary to define it again in the `to_hdf()` function, so the code is cleaner this way.

Leimeroth:
Looks nice, I will try it with some more complex notebooks

PR #77, opened 2021-02-10
Hotfix: Correct potential init for atomicrex

pmrv:
The issue is that even though `AbstractPotential` derives from `InputList`, it doesn't act like one, because it doesn't accept a list initializer in its init.  `InputList.from_hdf` depends on that possibility though.  I haven't tried it, but this should work.

However the better fix would be that `AbstractPotential` *doesn't* derive from `InputList` (since it is not intended to be used like one) and instead has a private member that *is* an `InputList`.  Then you can still use the list for HDF5 storage, but without breaking its protocol.

### DIFF
--- a/pyiron_contrib/atomistic/atomicrex/potential_factory.py
+++ b/pyiron_contrib/atomistic/atomicrex/potential_factory.py
@@ -23,18 +23,19 @@ class ARPotFactory(PyironFactory):
 
 
 class AbstractPotential(InputList):
-    def __init__(self, table_name="potential"):
-        super().__init__(table_name=table_name)
+    def __init__(self, init=None, table_name="potential"):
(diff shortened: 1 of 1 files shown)


jan-janssen (approved):
Works for me, thx

jan-janssen:
Still I guess we should discuss if it is possible to simplify the development of classes for pyiron - maybe on friday in the pyiron meeting. I guess @Leimeroth intention was to just derive from `InputList` and then the data storage is already taken care of and that is not yet the case.

Leimeroth:
> I guess @Leimeroth intention was to just derive from `InputList` and then the data storage is already taken care of and that is not yet the case.
Yes I derived so much from InputList because I like how they look in jupyter and because I hoped to avoid manually messing with the hdf5 input/output.

Issue #79, opened 2021-02-15
Resolve dependencies 


jan-janssen:
As we plan to use `pyiron_contrib` in the potential workshop, the dependency on fenics is increasing the dependencies of the overall docker image / environment. For the workshop we only need the `pyiron_contrib.atomistic` part, so we have to decide to either move the classes directly to `pyiron_atomistics` or we move the finite element part to `pyiron_experimental` or `pyiron_continuum` to have `pyiron_contrib` mainly for extensions of `pyiron_atomistics`. 

@liamhuber What is your take on this?

liamhuber:
Which atomistics classes do you need? If they're in good shape I don't object to moving them over to atomistics, but they should have unit tests if they don't already.

Contrib should definitely not become atomistic-centric. However, as long as we put a big enough disclaimer about the HDF5 storage (i.e. the lack thereof) I don't mind moving the FEM stuff over to continuum.

jan-janssen:
I started a corresponding repository https://github.com/pyiron/pyiron_continuum

PR #84, opened 2021-02-20
Modular requirements

jan-janssen:
Solves https://github.com/pyiron/pyiron_contrib/issues/79 

As pyiron_contrib contains different modules with rather different requirements, I suggest to make the dependencies which are specific to one module optional dependencies.


liamhuber (approved):
I had to google how `extras_require` works 😝 but no objections from my end 👍

PR #95, opened 2021-03-02
Improve TrainingContainer

sudarsan-surendralal:
Included a function to return all stored structures (also energies and forces) as a list

### DIFF
--- a/pyiron_contrib/atomistic/atomistics/job/trainingcontainer.py
+++ b/pyiron_contrib/atomistic/atomistics/job/trainingcontainer.py
@@ -29,10 +29,10 @@ Fe_bcc  ...
 """
 
 import pandas as pd
-
-from pyiron_atomistics import pyiron_to_ase
+from pyiron_atomistics import pyiron_to_ase, ase_to_pyiron
 from pyiron_atomistics.atomistics.structure.atoms import Atoms
 from pyiron_base import GenericJob
+
 
 class TrainingContainer(GenericJob):
     """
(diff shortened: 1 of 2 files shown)


pmrv (changes requested):
I like the filter functionality and the tests, but I'm not sure if  `get_data_list` gets that much more mileage than 
```python
p = tc.to_pandas()
p.atoms, p.energy, ...
```
Maybe we can just make a property for the filter function and use it in `to_pandas()`.   I'm a bit worried that converting to lists is going to take forever with large training sets.

If you want to keep it that's also ok with that, but then I'd change the name to `to_list` so that it is consistent with `to_pandas`.

EDIT: I realize that iterating over a dataframe is a bit cumbersome, but then I'd say we add a `__iter__` to the container that yields each structure, energy, forces and number of atoms.

sudarsan-surendralal:
> 
> 
> I like the filter functionality and the tests, but I'm not sure if `get_data_list` gets that much more mileage than
> 
> ```python
> p = tc.to_pandas()
> p.atoms, p.energy, ...
> ```

But is `p.atoms` giving us the pyiron atoms? I thought we still need to convert this which is what motivated this function

pmrv:
> > I like the filter functionality and the tests, but I'm not sure if `get_data_list` gets that much more mileage than
> > ```python
> > p = tc.to_pandas()
> > p.atoms, p.energy, ...
> > ```
> 
> But is `p.atoms` giving us the pyiron atoms? I thought we still need to convert this which is what motivated this function (and to avoid working with data frames which is useful for beginners)

Ah fair, no it will give the ase atoms.  That's mainly because our Atoms gives an error when pickled, but I couldn't look into it yet.  It's ok then for me, just change the name.

jan-janssen (approved):
Looks good to me - feel free to merge it, then we can create a new release for testing

PR #99, opened 2021-03-03
Runner

pmrv:
Adds a small class to fit RuNNer potentials, adapted from [here](https://github.com/jan-janssen/runner-example).  There's still a problem with writing out the input structures, I will contact Marius about this.

See also [here](https://github.com/pyiron/pyiron-resources/pull/35).


pmrv:
Before I merge this, I'll want to talk quickly in the meeting if we're going to need any kind of HDF functionality for the workshop or if this alright as it is.

jan-janssen (approved):
Looks good to me

PR #102, opened 2021-03-05
Simple atomistic

samwaseda:
And this probably marks my first PR to `pyiron_contrib`.

### DIFF
--- a/pyiron_contrib/atomistic/atomistics/job/simple_atomistic.py
+++ b/pyiron_contrib/atomistic/atomistics/job/simple_atomistic.py
@@ -1,0 +1,55 @@
+# coding: utf-8
+# Copyright (c) Max-Planck-Institut für Eisenforschung GmbH - Computational Materials Design (CM) Department
+# Distributed under the terms of "New BSD License", see the LICENSE file.
+
+from pyiron_atomistics.atomistics.job.interactive import GenericInteractive
+import numpy as np
+
+class SimpleAtomistic(GenericInteractive):  # Create a custom job class
+    def __init__(self, project, job_name):
+        super().__init__(project, job_name)
+        self.neigh = None
+        self.cutoff = 3
+        self.potential = None
+        self._neighbor_chemical_symbols = None
+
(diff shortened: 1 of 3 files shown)


pyiron_contrib/atomistic/atomistics/job/simple_atomistic.py
```diff
@@ -0,0 +1,66 @@
+# coding: utf-8
+# Copyright (c) Max-Planck-Institut für Eisenforschung GmbH - Computational Materials Design (CM) Department
+# Distributed under the terms of "New BSD License", see the LICENSE file.
+
+from pyiron_atomistics.atomistics.job.interactive import GenericInteractive
+import numpy as np
+
+class SimpleAtomistic(GenericInteractive):  # Create a custom job class
```
jan-janssen, line 8: It would be great to add a `to_hdf()` and `from_hdf()` method.

liamhuber: When you do this you might find it's convenient to change most (all?) fields to live on an `InputList`.

pyiron_contrib/atomistic/atomistics/job/simple_atomistic.py
```diff
@@ -0,0 +1,66 @@
...
+        self.neigh = None
+        self.potential = None
+        self._neighbor_chemical_symbols = None
+        self.uptodate = False
+
+    def interactive_initialize_interface(self):
+        pass
+
+    def validate_ready_for_run(self):
+        super(SimpleAtomistic, self).validate_ready_for_run()
```
liamhuber, line 20: nit: moderns `super` calls don't need the arguments (unless you have dual inheritance and want non-default resolution order)

pyiron_contrib/atomistic/atomistics/job/simple_atomistic.py
pmrv, line 8:
```suggestion
class SimpleAtomistic(GenericInteractive):
```

PR #119, opened 2021-03-22
Use DataContainer instead of the deprecated InputList

liamhuber:
Following the introduction of `DataContainer` which deprecates `InputList` in [pyiron_base PR 257](https://github.com/pyiron/pyiron_base/pull/257).

### DIFF
--- a/pyiron_contrib/atomistic/atomicrex/fit_properties.py
+++ b/pyiron_contrib/atomistic/atomicrex/fit_properties.py
@@ -1,9 +1,9 @@
 import xml.etree.ElementTree as ET
 
-from pyiron_base import InputList
+from pyiron_base import DataContainer
 
 
-class ARFitProperty(InputList):
+class ARFitProperty(DataContainer):
     """
     Class to describe properties that can be fitted using atomicrex.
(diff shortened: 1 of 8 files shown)


liamhuber:
@Leimeroth all the tests passed on my local machine, but I don't think I have any tests for atomicrex on my local machine. In principle, it should be a trivial re-name as all the old `InputList` functionality is captured in `DataContainer` (and more!) with the same interface. In practice, please check out the branch and make sure your workflows will still operate ok 😄

Leimeroth:
I ran a notebook and executing the jobs still works. They just can't be loaded from hdf anymore, but there were some bugs before as well. From my side its fine to merge and see how to fix the hdf issues later.

Leimeroth (approved):

pmrv:
> I ran a notebook and executing the jobs still works. They just can't be loaded from hdf anymore, but there were some bugs before as well. From my side its fine to merge and see how to fix the hdf issues later.

If the old jobs can't be loaded anymore that's a bug.  Can you create an issue on pyiron_base with the error?

PR #125, opened 2021-03-25
Move randspg

jan-janssen:


pyiron_contrib/atomistic/randspg/randspg.py
```diff
@@ -53,6 +55,26 @@ def list_structures(self):
         else:
             return []
 
+    def _executable_activate(self, enforce=False):
```
pmrv, line 58: Is overriding this still necessary after the move?

jan-janssen: You are right - this was part of the debugging.

PR #133, opened 2021-04-07
Rename atomistic to atomistics

### LINKED ISSUE #132: `atomistic` and `atomistics` coexisting?
In the `tests` folder I see `atomistics` and in the main folder there's `atomistics` inside `atomistic`. Is there particular logic in it?

liamhuber:
For consistency

To close #132


samwaseda (approved):
Maybe a bit unrelated, but does it have to be `atomistics.atomistics`? Not only it's not particularly elegant, but also it doesn't look like something crucial.

liamhuber:
> Maybe a bit unrelated, but does it have to be `atomistics.atomistics`? Not only it's not particularly elegant, but also it doesn't look like something crucial.

@samwaseda this is a fight to take up with `pyiron_atomistics.atomistics`. Here we just mimic the structure, but dropping the `pyiron_` prefix. E.g. before we had `pyiron_continuum` we had `pyiron_contrib.continuum`. I really like that setup more than I dislike nested atomistics 😛

PR #135, opened 2021-04-09
strain class implemented

samwaseda:


pyiron_contrib/atomistics/elasticity/bulk/strain.py
```diff
@@ -0,0 +1,21 @@
...
+def get_strain(structure, ref_structure, num_neighbors=None):
+    if num_neighbors is None:
+        cna = ref_structure.analyse.pyscal_cna_adaptive()
+        bulk = np.asarray([k for k in cna.keys()])[np.argmax([v for v in cna.values()])]
+        if bulk=='bcc':
+            num_neighbors = 8
+        elif bulk=='fcc' or bulk=='hcp':
+            num_neighbors = 12
+        else:
+            raise ValueError('Crystal structure not recognized')
```
liamhuber, line 12: My spidey-sense is tingling... what happens if someone tries to use this on a structure with a phase boundary? i.e. there are $N$ fcc atoms and $M<N$ bcc atoms (and $L<M<N$ 'other' atoms). What are the limitations here?

liamhuber (commented):
Necessary before `pyiron_contrib`:
- Docstring explaining all arguments
- PEP8
- `pyiron_contrib/atomistics/elasticity` should be a module, please add `__init__.py`

Necessary before moving to `pyiron_atomistic` (and so might as well get a jump-start here):
- Adding an example to the docstring
- Unit test

I also wonder how we will expose this to the user? Right now it's a public function buried down a module path. For `contrib` I'm totally fine with this. For the eventual transition over to `atomistics` we want to (a) be careful about what we expose as public methods as we're then on the hook to maintain backwards compatibility for the functionality (fine, it's cool functionality!) *and the interface* (then we want to be *super* careful about how we name things, which args to use, etc), and (b) how will we expose this functionality to the user?

For (b), in this case, I think it would be cool to have something like `struct.analyze.relative_strain(ref_struct)` as a wrapper. That's down the road, but I thought I'd throw out the interface.

Finally, I took a look also at `isotropic.py` and I am somewhat concerned about the [bus factor](https://en.wikipedia.org/wiki/Bus_factor) of this code. For `contrib` it is, again, totally fine, but for `atomistics` I would be more comfortable if there was some refactoring and more documentation of intent. E.g. AFAIK you're the only one in the department who's super comfortable with `einsum` -- I don't have any problem with you continuing to use it, but then the use should probably be wrapped in smaller sub-functions with very descriptive names so everyone else can read/maintain it easier.

pyiron_contrib/atomistics/elasticity/bulk/strain.py
samwaseda: Then they will probably get a monstrous strain in the wrong area XD. But that should be fine because nothing's gonna break down. So as long as the user knows what they are doing, they can simply superpose (?) two strain fields and get the right values from each result.

samwaseda:
> Necessary before `pyiron_contrib`:
> 
> * Docstring explaining all arguments
> * PEP8
> * `pyiron_contrib/atomistics/elasticity` should be a module, please add `__init__.py`
> 

Uh ok

pyiron_contrib/atomistics/elasticity/bulk/strain.py
liamhuber: Cool, fair enough. The docstring should make clear the underlying reason that `get_strain` works this way, e.g. something to the effect that it will only be sensible if the reference is one of fcc, bcc, hcp, and that it can be trivially spatially decomposed, or similar.

samwaseda:
> Finally, I took a look also at `isotropic.py` and I am somewhat concerned about the [bus factor](https://en.wikipedia.org/wiki/Bus_factor) of this code. For `contrib` it is, again, totally fine, but for `atomistics` I would be more comfortable if there was some refactoring and more documentation of intent. E.g. AFAIK you're the only one in the department who's super comfortable with `einsum` -- I don't have any problem with you continuing to use it, but then the use should probably be wrapped in smaller sub-functions with very descriptive names so everyone else can read/maintain it easier.

There, I have to say if you guys cannot come up with a solution to do linear algebraic operations in a 5-dim space (in this case `nijkl` for the second derivative of Green's function), I have to say I must stick to `einsum`. Let's take a random example from `isotropic.py`:

```
np.einsum('ij,nk,nl->nijkl', E, r, r)
```

This reads:

```
E_ij*r_nk*r_nl
```

As far as I know, the only other way to do the same is:

```
E[np.newaxis, :, :, np.newaxis, np.newaxis]*r[:, np.newaxis, np.newaxis, :, np.newaxis]*r[:, np.newaxis, np.newaxis, np.newaxis, :]
```

And I would VEEEEEEEEEEEEEERY surprised if someone prefers this notation. I'm ready to give you an overview at one of the pyiron meeting though if you want.

liamhuber:
> There, I have to say if you guys cannot come up with a solution to do linear algebraic operations in a 5-dim space (in this case `nijkl` for the second derivative of Green's function), I have to say I must stick to `einsum`. 

I have absolutely zero problem with this.

What I want is that the reader is able to understand the basic mathematical operations here without knowing anything about `einsum`'s *syntax* -- or `np.tensordot` or any other non-standard syntax for that matter. Just about the only exception I would make here is that `np.einsum('ij,ik->jk', ref_frame, ref_frame)` can be replaced with `ref_frame.T @ ref_frame`.

My suggestion to break free of syntax dependence is to wrap the math in well-named functions. Unfortunately, I don't really know much about elasticity and what you're doing here is not totally clear to me from the `einsums` but it might look something like this:

Now:
```python
D = np.einsum('ij,ik->jk', ref_frame, ref_frame)
D = np.linalg.inv(D)
J = np.einsum('nij,nik->njk', ref_frame[indices], neigh.vecs)
J = np.einsum('ij,njk->nik', D, J)
return 0.5*(np.einsum('nij,nkj->nik', J, J)-np.eye(3))
```

Want:
```
D = _inverse_lattice_constants(ref_frame)
J = _compare_displacements(ref_frame[indices], neigh.vecs)
foo = _contract_something_maybe(D, J)
return _symmetrize(foo)

def _inverse_lattice_constants(reference_neighbors):
    """Just a diagonal matrix with inverse cell dimensions or what?"""
    return np.linalg.inv(reference_neighbors.T @ reference_neighbors)

def _compare_displacements(reference_vecs, actual_vecs):
    return np.einsum('nij,nik->njk', reference_vecs, actual_vecs)

def _contract_something(a_tensor, another_one)
    return np.einsum('ij,njk->nik', a_tensor, another_one)

def _symmetrize(tensor):
    return 0.5*(np.einsum('nij,nkj->nik', J, J)-np.eye(3))
```

Again, I can't emphasize enough that the actual names there are crap, but I hope the idea comes across despite this.

samwaseda:
> Want:
> 
> ```python
>         ...
>         D = _inverse_lattice_constants(ref_frame)
>         J = _compare_displacements(ref_frame[indices], neigh.vecs)
>         foo = _contract_something_maybe(D, J)
>     return _symmetrize(foo)
> 

I generally agree with your idea, but here the problem is that `D` has absolutely no physical meaning and `J` doesn't really have a name (I called it rotation matrix recently, which @alitehranch correctly pointed out was an absolutely wrong name). I could call the whole thing `get_lagrangian_tensor` or so. But then, I'm not sure if it's particularly meaningful, because I don't think there's anything other than Lagrangian tensor anyway in atomistic simulations...

liamhuber:
>  but here the problem is that `D` has absolutely no physical meaning 

I disagree. I checked fcc, hcp, and bcc -- `D` was always diagonal, and the two cubic structures just gave $x.I$, while hcp roke the z-symmetry and the (2, 2) entry was unique. From this I'm extremely confident that there is physical meaning here. But like I said, I don't really understand what you're doing, so it must be you that determines the details and describes them.

> and `J` doesn't really have a name (I called it rotation matrix recently, which @alitehranch correctly pointed out was an absolutely wrong name). I could call the whole thing `get_lagrangian_tensor` or so. But then, I'm not sure if it's particularly meaningful, because I don't think there's anything other than Lagrangian tensor anyway in atomistic simulations...

I have faith that you will come up with something reasonably descriptive. At least as a start, not overloading and recursively defining the variable is a good start -- right now `J = f(); J = G(J)`...

liamhuber:
I was thinking about explaining *why* I am so hung up on having this more readable. A big part of it is thinking ahead to the person who will eventually come to this code and need to extend it, and what they are starting with in that job.

That was the tangential motivation, but I think I stumbled on a practical improvement 😛 Since this only works for FCC, HCP, and BCC right now any way, why even require a reference structure? Why not
- CNA encode all the atoms
- Build separate reference structures for each CNA tag we accommodate 
- Make separate local strain analyses for each site based on its CNA tag
- Combine the separate analyses back together with NaN values for the neglected tags

Would this work? It would greatly improve flexibility and make it more robust to user error because you'd just get a NaN if you tried using it somewhere the local environment wasn't well treated.

liamhuber:
> Would this work? 

nevermind, I guess at least you need also the lattice constants for each appropriate tag, and it would still fail if you had, say, two incomensurate materials with the same crystal structure together. Not so easy.

samwaseda:
> > Would this work?
> 
> nevermind, I guess at least you need also the lattice constants for each appropriate tag, and it would still fail if you had, say, two incomensurate materials with the same crystal structure together. Not so easy.

I was about to write this XD

samwaseda:
So I can refer this [article] in the code (and also change the variable names) if you want more physics clarification. If that's not satisfactory, I'll move it to `pyiron_mpie`.

liamhuber:
> So I can refer to this [article] in the code (and also change the variable names) if you want more physics clarification. If that's not satisfactory, I'll move it to `pyiron_mpie`.

Article link didn't show up.

And `contrib` is ok. I only had three conditions for contrib (above). I would be uncomfortable transferring it to atomistics in the current state though.

samwaseda:
> Article link didn't show up.

Sorry I updated my comment ([link also here](http://li.mit.edu/A/Graphics/A/annotate_atomic_strain/Doc/main.pdf))

> 
> And `contrib` is ok. I only had three conditions for contrib (above). I would be uncomfortable transferring it to atomistics in the current state though.

So if it's mostly about docstring I can probably make it better. But if it's more about `einsum` or giving variables meaningful names, then I'd be a bit in trouble, because for me `einsum` is still the most transparent method when it comes to tensor multiplications of dim>3 and the variables don't really have a meaning. Let's come back to your example:

```python
        ...
        D = _inverse_lattice_constants(ref_frame)
        ...

def _inverse_lattice_constants(reference_neighbors):
    """Just a diagonal matrix with inverse cell dimensions or what?"""
    return np.linalg.inv(reference_neighbors.T @ reference_neighbors)
```

(By the way this is not a diagonal matrix)

I really appreciate your effort of giving it a name, but I don't see why this is easier to understand than this:

```python
    D = np.einsum('ij,ik->jk', ref_frame, ref_frame)
    D = np.linalg.inv(D)
```

I'm actually even a bit concerned that by creating new functions we might end up a) forcing the user to go back and forth in the code and b) misleading them by showing a subjective function name, or confusing them with a totally esoteric name (e.g. `_inverse_sum_frame_outer_products` in this case).

pmrv:
> I'm actually even a bit concerned that by creating new functions we might end up a) forcing the user to go back and forth in the code and b) misleading them by showing a subjective function name, or confusing them with a totally esoteric name (e.g. `_inverse_sum_frame_outer_products` in this case).

In the case you can define nested functions in the method in the order that you'll be using them.  I think using the einsum's is fine, but there should be documentation (as functions or comments, idc) on what they are implementing.

liamhuber:
> I think using the einsum's is fine, but there should be documentation (as functions or comments, idc) on what they are implementing.

> I'm actually even a bit concerned that by creating new functions...

Yeah, I'm definitely not totally hung up that it needs to be in the form of new functions. As @pmrv suggests, inline comments may also be a good way to go here. Sometimes simply introducing something as a well named variable is sufficient. 

Critical is that there is more documentation, not just at the user level (i.e. a good docstring) but at the developer level (i.e. in how the raw code reads). Right now the intentions, scope, physics, and limitations of this code are not at all clear. For instance, right now it's called `get_strain`, but what type? I am really not an expert here, but my understanding is there are many [representations](https://en.wikipedia.org/wiki/Finite_strain_theory). 

Suppose three years from now a new developer wants to do something in this direction, so they come into the code base to see what is already available and how they can extend it. My feeling is that right now there is a lot of work involved to get up to speed. At the lowest level, the reference for the equations should be given. At a higher level, the equation using the variables should be shown. At the highest level, the steps of the solution should be described (comments/variable names/refactored functions) so that a non-expert can read and understand the mathematics/physics.

For `contrib` we don't make promises about backwards compatibility so I don't mind merging this even if we're not at such an ideal. But all of us have (relatively) short term contracts. Our "bus factor" is through the roof. IMO readability/maintainability/extensibility needs to be *extremely* high on our priority list. So for `contrib` w/e, but while we're here and have this code at the front of our minds is a good time to make steps towards having it `atomistics`-ready.

liamhuber:
> ...the variables don't really have a meaning.

> ...misleading them by showing a subjective function name, or confusing them with a totally esoteric name (e.g. `_inverse_sum_frame_outer_products` in this case).

I still really struggle to accept this. The final result has physical meaning, and we do some sensible operations along the way. How is it impossible to come up with a decent description of what is being done?

In this particular case I am 100% convinced that `outer(ref_neighbor_vecs, ref_neighbor_vecs)` has physical meaning. I happened to be diagonal for all the perfect crystals, but you say isn't generally diagonal. HCP broke the z-axis symmetry. This smells to high-heaven of physical meaning.

muh-hassani:
I have not dug deep into the code. But from the comments, if the strain evaluated can not be applied to a general atomistic structure, its application would be limited. I have some experience with the evaluation of atomic strain using coarse-grained atomic displacement. 
equation
where `\vec{u}_j` is the displacement of atom `j`; `\phi` is a Gaussian function.`N` is the number of atom in the neighborhood. Since `\phi` is a continuous function, you can apply gradient to it to calculate the strain at the site of any atom.   
WIth this method, I guess you can avoid the limitations for a certain atomic structure and boundary condition.

pmrv:
> > ...the variables don't really have a meaning.
> 
> > ...misleading them by showing a subjective function name, or confusing them with a totally esoteric name (e.g. `_inverse_sum_frame_outer_products` in this case).
> 
> I still really struggle to accept this. The final result has physical meaning, and we do some sensible operations along the way. How is it impossible to come up with a decent description of what is being done?
> 
> In this particular case I am 100% convinced that `outer(ref_neighbor_vecs, ref_neighbor_vecs)` has physical meaning. I happened to be diagonal for all the perfect crystals, but you say isn't generally diagonal. HCP broke the z-axis symmetry. This smells to high-heaven of physical meaning.

In a way that's the first of MTP's moment tensors, so make of that what you will. :P

samwaseda:
I implemented it in pyiron_mpie in the end ([here](https://github.com/eisenforschung/pyiron_mpie/pull/45)), but since this is a recurring issue maybe I still say how I see the problem with a simplified version of what I'm doing, namely linear least square fit.

So the linear least square method is a method which tries to find a function `f(x) = a*x + b` which fits a set of points `(x_i, y_i)`, right? In short what is done is to minimise the function:

```
err(a, b) = sum_i (y_i - a*x_i - b)
```

And then `a` and `b` are given by:

```
a = (N*(sum_i x_i*y_i) - (sum_i x_i)*(sum_i y_i))/(N*(sum_i x_i^2)-(sum_i x_i^2))
b = ((sum_i y_i)*(sum_i x_i^2) - (sum_i x_i)*(sum_i x_i*y_i))/(N*(sum_i x_i^2)-(sum_i x_i^2))
```

Now what you guys are asking me to do is essentially to find names for `(N*(sum_i x_i*y_i) - (sum_i x_i)*(sum_i y_i))` and ` ((sum_i y_i)*(sum_i x_i^2) - (sum_i x_i)*(sum_i x_i*y_i))`. Sure, they might have physical meanings and they might nave names. But ultimately what I'm doing is to minimise `err(a, b)` and find `a` and `b`. And while `a` and `b` certainly should have names, I would even claim that the rest should **not** even have names, because even they can be referred to as something existing, it doesn't mean that it's the universal usage of the name (in which case the user is only misguided), or the name is so uncommon that it helps no one to understand what it actually means.

What I'm trying to say in this example is not how difficult it was to give a name to the term `D` I had in the code, but I'm much more trying to say that in math there are simply painful operations which have to be done until something meaningful comes out. In this case, I prefer to use something that for sure cannot be understood (e.g. just `D`), rather than trying to find something superficially descriptive, so that if the code has to be modified, the user **must** look at the actual content of the code.

samwaseda:
And I removed all elasticity codes from `pyiron_contrib`, because linear elasticity is often a high dimensional problem, which forces me to use `einsum`, which seems to cause more trouble to people than profit.

liamhuber:
> And I removed all elasticity codes from `pyiron_contrib`, because linear elasticity is often a high dimensional problem, which forces me to use `einsum`, which seems to cause more trouble to people than profit.

That's certainly your prerogative. To reiterate though, I don't have any intrinsic problem with `einsum` itself, and agree with you that I'm not aware of any clearer syntax for dimensions >3.

liamhuber:
> Now what you guys are asking me to do is essentially to find names for `(N*(sum_i x_i*y_i) - (sum_i x_i)*(sum_i y_i))`.

I mean, isn't this directly proportional to  the [covariance](https://en.wikipedia.org/wiki/Covariance), `N^2 * covariance(x, y)`? 

For the other numerator I don't so easily see a good description.

> I prefer to use something that for sure cannot be understood

I guess I just really struggle with intentional obfuscation. I see your argument in favour of it, but it seems like a "desperate times/desperate measures" approach.

samwaseda:
> 
> I guess I just really struggle with intentional obfuscation. I see your argument in favour of it, but it seems like a "desperate times/desperate measures" approach.

There are simply mathematical equations which are inevitably long. The one that you complained about first was the second derivative of the Green's function. It's not because it consists of several parts, and indeed the Green's function itself is relatively short, but for the mere fact that we need the second derivative of it, it inevitably becomes long. And here in this strain class as well, the error function itself is short, but for the mere fact that we try to give an analytical solution it inevitably becomes long, again not because it consists of several parts. If pyiron doesn't want to accept this only because it becomes difficult to maintain, then we'll soon limits in some of the classical theories, like linear elasticity as we can see here in this example.

liamhuber:
> If pyiron doesn't want to accept this only because it becomes difficult to maintain, then we'll soon limits in some of the classical theories, like linear elasticity as we can see here in this example.

I think it would be an absolute shame to exclude functionality like you show here. As pyiron expands, it's my opinion we'll *need* such tools.

I also want to reiterate -- and really for the final time, because I counted and I've already been explicit about this five times in this thread -- **I *do not object* to merging difficult-to-maintain chunks of mathematics into `pyiron_contrib`**, whether they contain `einsum` or not. I don't think anyone else is objecting to it either.

My original three concerns for merging into `contrib` are the same, and the crux is really *user level* documentation, i.e. a docstring explaining input/output and maybe discussing the limitations of the function.

> Necessary before pyiron_contrib:
> 
> - Docstring explaining all arguments
> - PEP8
> - pyiron_contrib/atomistics/elasticity should be a module, please add __init__.py

For merging into `pyiron_atomistics` I have to continue to insist that we put a high value on readability/maintainability/extensibility -- how can I do otherwise? Definitely the degree to which we *succeed* at those three priorities will vary for different functionality. I absolutely agree that analytic models from classical theories are challenging to make readable. We certainly aren't bound to breaking them into small sub-functions; as Marvin suggested in-line comments may play the key role, and as you suggested we can link to the source paper (and reproduce the equations in text/latex?) and synchronize variable names.

The only hill that I'm willing to die on here is that we mustn't give up entirely on readability. It should simply always be an important objective. How well we succeed at it will vary depending on the content we're creating, and that's OK.

PR #140, opened 2021-04-13
hdf load workaround for 2 classes with same name

Leimeroth:
There still was an issue with loading jobs from hdf5 if they used the gaussian function. I assume the reason is the same class name for 2 different classes.


pmrv:
If I have this on my radar correctly, classes are saved by there full import path + name, so I'm not sure I understand the issue.  We should talk about this next week to see if I miss something.

PR #148, opened 2021-04-23
Minimal structure and structure list

Leimeroth:
Completely rewrote how structures and fit properties are stored for atomicrex jobs.
It looks much worse than the DataContainer based storage, but this makes writing (and loading) hdf files 50-100 times faster on our system and reduces the file size by > factor 10 for some of my jobs.
Right now it is a bit inconvenient to use because it is necessary to give the number of structures and total number of atoms manually, which allows me to preallocate numpy arrays with the necessary size. I am open to suggestions how to improve this without massive slowdowns cause by repeated allocations.
Will add some documentation in the next few days.


pyiron_contrib/atomistics/atomicrex/structure_list.py
```diff
@@ -8,64 +8,316 @@
...
+
+    def add_structure(self, structure, identifier):
+        # len of structure to index into the initialized arrays
+        n = len(structure)
+        i = self.current_atom_index + n
+
+        self.len_current_struct[self.current_structure_index] = n
+        self.symbols[self.current_atom_index:i] = np.array(structure.symbols)
+        self.positions[self.current_atom_index:i] = structure.positions
+        self.cells[self.current_atom_index:i] = structure.cell.array
```
pmrv, line 63:
```suggestion
        self.cells[self.current_structure_index] = structure.cell.array
```

pmrv (changes requested):
The new "flattened" classes look good to me.

On the topic of pre-allocating, I did some small benchmarks like this

```python 
import numpy as np

N = 100_000_000
M =     100_000
bigly = np.random.rand(N)


def pre():
    # ~1.3s
    a = np.empty(N)
    for i in range(N // M):
        a[i * M : (i+1) * M] = bigly[i * M : (i+1) * M]
    return a

def grw():
    # 400-700ms
    N_alloc = M
    N_sofar = 0
    a = np.empty(N_alloc)
    for i in range(N // M):
        while N_alloc < N_sofar + M:
            N_alloc *= 2
            a.resize(N_alloc)
        a[i * M : (i+1) * M] = bigly[i * M : (i+1) * M]
        N_sofar += M
    a.resize(N)
    return a
```

`pre()` allocates full storage and then fills step by step from `bigly` as the current code would do.  `grw()` on the other hand resizes the allocated storage dynamically as it copies new data into it.   I then benched it with the ipython `%timeit` magic.  For some reason `grw()` is even faster than `pre()`.  I'm not sure if we can expect this to translate to our more complicated case, but at least dynamically resizing shouldn't be a performance issue.  

Given that I think that `FlattenedStructureContainer` and friends should be able to grow as they go.  The current code can stay since there might be cases where you know the number of structures ahead of time, but it should be expanded to resize when new structures are added and shrunk to the minimum possible when written to HDF5.

pyiron_contrib/atomistics/atomicrex/structure_list.py
Leimeroth: how did this even work and not throw an index error

Leimeroth:
> 
> On the topic of pre-allocating, I did some small benchmarks like this
> 

I just copy pasted your code and ran it. Dynamically resizing takes more than twice as long for me, which is better than expected but still much worse than pre allocating. I am quite confused how dynamically resizing can be faster on your computer.
pre(): 107 ms ± 878 µs
grw(): 232 ms ± 4.09
I also tried with another 0 on N and M and ran it several times, always takes about 2.3 times as long.

pmrv:
So if I change `M` to be larger or smaller `pre` becomes faster than `grw`, but on my machine it stays within 10%, I don't know what's going on there.  On our cluster login node, `pre` seems to be around 50% faster.  

In any case right now we need to store N x M x (3 + 1) numbers, where N structures, M atoms per structure and for each we have positions (3) and types (1).  There's also cells and other information per structure, but that's O(N).  I would think that we can expect N ~ 10'000 - 100'000 and M ~ 10 - 1'000, which according to our benchmarks above is < 1s.  Even if we decide later to store much more information we are still below 10s.  Given that it currently takes much longer, it seems unlikely to me that we would bottleneck in this part of the code.  I would therefore suggest that we keep the current code that pre-allocates the arrays, but put default values there and allow the arrays to grow if more structures are added.  That way users can benefit from the fast pre-allocation where it helps them, but also have a flexible interface in case they don't know the number of structures a-priori.  If you're busy I can patch that in on Monday.

Leimeroth:
I changed the re-allocations and num structures from >= conditions to > conditions, because the >= broke my code. This shouldn't cause any issues imo, but please see if you find something @pmrv .

Leimeroth (commented):
Side note:  Using np.resize() instead of the resize method takes about 1.5 to 2 times longer in your simple test on my computer.

Leimeroth:
Locally works for me aswell

pmrv:
@Leimeroth I'd like to merge this into master today or tomorrow to start the work of getting the flattened structure list into atomistics.  I'm working on the resize problem for now, but any other blockers for you?

Leimeroth:
From my side it's fine, just didn't want to merge with failing tests

PR #153, opened 2021-04-26
Add convenience method to recreate added structures from container

pmrv:
Adds a `get_structure()` method as Jörg suggested on Friday.

### DIFF
--- a/pyiron_contrib/atomistics/atomicrex/structure_list.py
+++ b/pyiron_contrib/atomistics/atomicrex/structure_list.py
@@ -103,33 +103,32 @@ class FlattenedStructureContainer:
             self.identifiers = hdf_s_lst["identifiers"].astype(np.dtype("U20"))
             self.len_current_struct = hdf_s_lst["len_current_struct"]
 
-    def get_structure(self, iteration_step=-1, wrap=False):
+    def get_structure(self, index_or_identifier=-1):
         """
         Create pyiron Atoms object from saved arrays.
 
         Args:
-            iteration_step (int, str): if str then must one of the identifiers given to :method:`.add_structure`
-                                       otherwise must be and integer in range [-N, N) where N the number of structures
-                                       stored in the container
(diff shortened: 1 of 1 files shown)


Leimeroth (commented):
I like that structures can be obtained using the identifier or the index, but I think calling it iteration_step only makes sense for MD simulations, but for those the FlattenedStructureContainer is still a bit overkill because the number of atoms in each step and, if the same order is kept, the symbols stay constant. Are there some plans to use this or a similar class for storing MD calculations?
A name like index_or_identifier or splitting it into 2 different arguments would be more self explanatory imo for the FlattenedStructureContainer.
Also E could be obtained using I + self.len_current_structure(iteration_step), which is easier I guess.

pmrv:
> I like that structures can be obtained using the identifier or the index, but I think calling it iteration_step only makes sense for MD simulations, but for those the FlattenedStructureContainer is still a bit overkill because the number of atoms in each step and, if the same order is kept, the symbols stay constant. Are there some plans to use this or a similar class for storing MD calculations?
> A name like index_or_identifier or splitting it into 2 different arguments would be more self explanatory imo for the FlattenedStructureContainer.

The arguments I named as per convention in the other atomistic jobs.  It's not ideal here, I agree.  I'll change it for now and then later see how to harmonize it with [this](https://github.com/pyiron/pyiron_atomistics/pull/167) later.


> Also E could be obtained using I + self.len_current_structure(iteration_step), which is easier I guess.

Agree, didn't catch that.

pmrv:
> Are there some plans to use this or a similar class for storing MD calculations?

I would like this class to completely replace `StructureContainer` from pyiron_atomistics, so also structures from MD will end up in there at some point.  I also want this to work in the context of (semi-) grandcanonical MC, so I'd rather not put assumptions about number/types of atoms during a run in the class, too.

Leimeroth (approved):
I think you can merge it.

PR #154, opened 2021-04-27
Resize flattened container on the fly

pmrv:
As per our discussion in #148.

### DIFF
--- a/pyiron_contrib/atomistics/atomicrex/structure_list.py
+++ b/pyiron_contrib/atomistics/atomicrex/structure_list.py
@@ -22,12 +22,21 @@ from pyiron_contrib.atomistics.atomicrex.fit_properties import FlattenedARProper
 class FlattenedStructureContainer:
     """
     Class that can write and read lots of structures from and to hdf quickly.
-    This is done by storing positions, cells, etc. into large arrays instead of writing every structure into a new group.
-    Structures are stored together with an identifier that should be unique.
-    The class has to be initialized with the number of structures and the total number of atoms in all structures.
-    This allows to preallocate the necessary arrays and therefore prevents reallocations.
+
+    This is done by storing positions, cells, etc. into large arrays instead of writing every structure into a new
+    group.  Structures are stored together with an identifier that should be unique.  The class can be initialized with
+    the number of structures and the total number of atoms in all structures, but re-allocates memory as necessary when
+    more (or larger) structures are added than initially anticipated.
     """
+
     def __init__(self, num_structures=0, num_atoms=0):
+        """
+        Create new structure container.
+
+        Args:
+            num_structures (int): pre-allocation for per structure arrays
+            num_atoms (int): pre-allocation for per atoms arrays
+        """
         self._num_structures_alloc = num_structures
         self._num_atoms_alloc = num_atoms
         # store the starting index for properties with unknown length
--- a/pyiron_contrib/atomistics/atomicrex/structure_list.py
+++ b/pyiron_contrib/atomistics/atomicrex/structure_list.py
@@ -53,10 +53,10 @@ class FlattenedStructureContainer:
     def _init_arrays(self):
         self.symbols = np.full(self._num_atoms_alloc, "XX", dtype=np.dtype("U2"))# 2 character unicode array for chemical symbols
         self.positions = np.empty((self._num_atoms_alloc, 3))
-        self.cells = np.empty((self.__num_structures_alloc_alloc, 3, 3))
-        self.start_indices = np.empty(self.__num_structures_alloc_alloc, dtype=np.int32)
-        self.len_current_struct = np.empty(self.__num_structures_alloc_alloc, dtype=np.int32)
-        self.identifiers = np.empty(self.__num_structures_alloc_alloc, dtype=np.dtype("U20"))
+        self.cells = np.empty((self.__num_structures_alloc, 3, 3))
+        self.start_indices = np.empty(self.__num_structures_alloc, dtype=np.int32)
+        self.len_current_struct = np.empty(self.__num_structures_alloc, dtype=np.int32)
+        self.identifiers = np.empty(self.__num_structures_alloc, dtype=np.dtype("U20"))
 
     def _resize_atoms(self, new):
         self._num_atoms_alloc = new
--- a/pyiron_contrib/atomistics/atomicrex/structure_list.py
+++ b/pyiron_contrib/atomistics/atomicrex/structure_list.py
@@ -37,12 +37,13 @@ class FlattenedStructureContainer:
             num_structures (int): pre-allocation for per structure arrays
             num_atoms (int): pre-allocation for per atoms arrays
         """
-        self._num_structures_alloc = num_structures
-        self._num_atoms_alloc = num_atoms
+        # tracks allocated versed as yet used number of structures/atoms
+        self._num_structures_alloc = self.num_structures = num_structures
+        self._num_atoms_alloc = self.num_atoms = num_atoms
         # store the starting index for properties with unknown length
-        self.num_atoms = 0
+        self.current_atom_index = 0
         # store the index for properties of known size, stored at the same index as the structure
-        self.num_structures = 0
+        self.current_structure_index = 0
         # Also store indices of structure recently added
         self.prev_structure_index = 0
         self.prev_atom_index = 0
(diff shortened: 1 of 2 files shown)


pyiron_contrib/atomistics/atomicrex/structure_list.py
```diff
@@ -26,33 +26,49 @@ class FlattenedStructureContainer:
...
     This allows to preallocate the necessary arrays and therefore prevents reallocations.
-
-    TODO: Alternatively preallocate arrays with large default values (f.e. 1e6 atoms) then use resize to shrink them?
-    """    
+    """
     def __init__(self, num_structures=0, num_atoms=0):
-        self.num_structures = num_structures
-        self.num_atoms = num_atoms
+        self._num_structures_alloc = num_structures
+        self._num_atoms_alloc = num_atoms
```
Leimeroth, line 32: I use num_structures and num_atoms in some other places in the atomicrex code, f.e. to iterate over the structures, so I wouldn't change their name.

pyiron_contrib/atomistics/atomicrex/structure_list.py
```diff
@@ -75,24 +91,25 @@ def add_structure(self, structure, identifier):
 
 
     def to_hdf(self, hdf, group_name="flattened_structures"):
-        if self.symbols[-1] == "XX":
-            raise ValueError("Initialized arrays are not filled with values")
+        # truncate arrays to necessary size before writing
+        self._resize_atoms(self.current_atom_index)
```
Leimeroth, line 95: Resizing to current_atom_index truncates the positions of the last structure I think. This should probably be current_atom_index + self.len_current_structure[self.current_structure_index]

pyiron_contrib/atomistics/atomicrex/structure_list.py
Leimeroth: Maybe it is a good idea to have _num_structures_alloc for internal reallocations and num_structures for other uses like indexing and resizing before storing to hdf. When reading from hdf we allocate using the actual numbers and when structures are added they can be used as internal helper variables

pyiron_contrib/atomistics/atomicrex/structure_list.py
```diff
@@ -26,33 +26,49 @@ class FlattenedStructureContainer:
...
-        self.start_indices = np.empty(self.num_structures, dtype=np.int32)
-        self.len_current_struct = np.empty(self.num_structures, dtype=np.int32)
-        self.identifiers = np.empty(self.num_atoms, dtype=np.dtype("U20"))
-
+        self.symbols = np.full(self._num_atoms_alloc, "XX", dtype=np.dtype("U2"))# 2 character unicode array for chemical symbols
+        self.positions = np.empty((self._num_atoms_alloc, 3))
+        self.cells = np.empty((self.__num_structures_alloc_alloc, 3, 3))
+        self.start_indices = np.empty(self.__num_structures_alloc_alloc, dtype=np.int32)
+        self.len_current_struct = np.empty(self.__num_structures_alloc_alloc, dtype=np.int32)
+        self.identifiers = np.empty(self.__num_structures_alloc_alloc, dtype=np.dtype("U20"))
```
Leimeroth, line 50: I still would like to keep num_structures, but if we decide otherwise it should be _num_structures_alloc and not __num_structures_alloc_alloc I guess.

pyiron_contrib/atomistics/atomicrex/structure_list.py
```diff
@@ -26,33 +26,49 @@ class FlattenedStructureContainer:
...
+        self.len_current_struct.resize(new)
+        self.identifiers.resize(new)
 
     def add_structure(self, structure, identifier):
+
+        new_atoms = self.current_atom_index + len(structure)
+        if new_atoms >= self._num_atoms_alloc:
+            self._resize_atoms(max(new_atoms, self._num_atoms_alloc * 2))
+        if self.current_atom_index + 1 >= self._num_structures_alloc:
+            self._resize_structures(self._num_structures_alloc * 2)
```
Leimeroth, line 70: For atomicrex it is probably unnecessary but I wonder if we should consider cases where doubling the array size leads to arrays greater than memory?

pyiron_contrib/atomistics/atomicrex/structure_list.py
pmrv: Fair enough, what about I rename `current_structure_index -> num_structures` and `current_atom_index -> num_atoms`?

pyiron_contrib/atomistics/atomicrex/structure_list.py
pmrv: :S

pyiron_contrib/atomistics/atomicrex/structure_list.py
pmrv: I'm not super worried that we run into that, but I'll check how to get that limit.

pyiron_contrib/atomistics/atomicrex/structure_list.py
Leimeroth: Nevermind current_atom_index is updated with add_structure so this should work.

pyiron_contrib/atomistics/atomicrex/structure_list.py
Leimeroth: current_structure_index and current_atom_index are used to store the position in the structure and positions array after adding a structure, while num_structures and num_atoms store the total number of structures resp. atoms in the container and I think both are necessary for the functionality.

pmrv: But immediately after adding a structure the number of structures is the same as the index for the next structure or am I missing something?  Same for the atoms.

I see the scenarios:
1. immediately after creation of the container `num_structures=0`, `_num_structures_alloc=100`.  There's no structures inside yet and the first structures should be written at index 0, `current_structure_index=0`
2. after some structures have been added `num_structures=n`, `_num_structures_alloc=m>n`. The first `n` values of say, `self.cells` are taken, so the next one should go to index `n`, so `current_structure_index=n`.

pyiron_contrib/atomistics/atomicrex/structure_list.py
pmrv: I can't find a cross platform solution without third-party packages and I don't think it's worth extra dependencies.

pyiron_contrib/atomistics/atomicrex/structure_list.py
Leimeroth: I can't think of a scenario where there wouldn't be an easy workaround that needs only 2 of those 3,
but considering clarity of names and _ naming convention I still think we should have all 3.
Then num_structures allows to iterate over all the structures, current_structure can be used to index into the array
and _num_structures_alloc is used for internal resizing and the cost to store 2 more ints is marginal cosidering that we store arrays with possibly millions of entries.
This also allows for some possibly nice features, f.e. we could set current_structure_index to the corresponding index when using get_structure(), which then allows to use it in arrays that store additional data.
For example in the atomicrex class I could index into my fit_property arrays using it then.

Alternatively I would leave out the _num_structures_alloc and use the num_structures for internal resizing, but then we need to  be super careful to always set it back to the correct number after reallocations, which probably gets much worse if this is used together with other flattened properties, like I do to store the fit_properties.

pmrv: Fair enough, if it makes other things easier, I'll add it back in.

pyiron_contrib/atomistics/atomicrex/structure_list.py
Leimeroth: > I don't think it's worth extra dependencies.
I agree. Also users that want to fit millions of structures or something similar should be aware of possible memory problems.

pmrv:
From my side this is ready, anything else you'd want me to change?

Leimeroth:
What do you think about the current_structure_index? 

> I can't think of a scenario where there wouldn't be an easy workaround that needs only 2 of those 3,
> but considering clarity of names and _ naming convention I still think we should have all 3.
> Then num_structures allows to iterate over all the structures, current_structure can be used to index into the array
> and _num_structures_alloc is used for internal resizing and the cost to store 2 more ints is marginal cosidering that we store arrays with possibly millions of entries.
> This also allows for some possibly nice features, f.e. we could set current_structure_index to the corresponding index when using get_structure(), which then allows to use it in arrays that store additional data.
> For example in the atomicrex class I could index into my fit_property arrays using it then.

Also what do you think is the smartest way to use the resizing capability in other classes that use the structure container, but extend it? I would additionally define the resizing functions in my ARStructureContainer  for the fit properties, but that feels like a lot of doubled code, so maybe there is some better way?

pmrv:
> What do you think about the current_structure_index?

Forgot to push the change. :/

> Also what do you think is the smartest way to use the resizing capability in other classes that use the structure container, but extend it? I would additionally define the resizing functions in my ARStructureContainer for the fit properties, but that feels like a lot of doubled code, so maybe there is some better way?

The way I want to do that is by adding a way to "register" new per atom or per structure properties with the structure container, that would be useful for the `TrainingContainer` and also for the your fitting properties.  I have an idea how to do that, but I want to merge this first with the `pyiron_atomistics.StructureContainer` to see how this pans out.  A brief sketch could be

```python
class StructureContainer:
  def register(self, propname, per_structure_or_atom):
    if per_structure_or_atom == 'atom':
      self._atoms[propname] = np.empty(self._num_atoms_alloc)
    else:
      self._structures[propname] = np.empty(self._num_structures_alloc)

  def resize_atoms(new_atoms):
    for p in self._atoms:
      self._atoms[p].resize(...)

  def resize_structures(new_structures):
    for p in self._structures:
      self._structures[p].resize(...)

  def to_hdf(self, hdf):
    for p in self._atoms.keys() + self._structures.keys():
      hdf[p] = ...
```

Leimeroth (approved):

PR #163, opened 2021-05-12
Nested reference fixes for TILD

liamhuber:
`DecoupledOscillators` are now working when given as a reference job to a *serial* TILD job. Parallel is still broken somehow.


pyiron_contrib/protocol/jobs/decoupled_oscillators.py
```diff
@@ -124,42 +160,35 @@ def _set_base_structure(self):
         self._base_atom_ids = np.delete(np.arange(len(self.input.structure)).astype(int),
                                         self.input.oscillators_id_list)
 
-    def _create_base_job(self, pr=None, name=None):
+    @property
+    def _base_name(self):
+        return self.job_name + '__base'
```
raynol-dsouza, line 165: I'd rather just want to leave the extension as '_base' than '__base', unless there's a specific reason for the dunder?

liamhuber: I always dunder auto generated names. It's not a guarantee but adds a level of safety. Ie imagine a project where Someone creatively names this job "job" and then later in the project in an equally impressive bout of creativity calls something else "job_base". So my rule of thumb is to dunder auto names

pyiron_contrib/protocol/jobs/decoupled_oscillators.py
```diff
@@ -253,18 +282,12 @@ def from_hdf(self, hdf=None, group_name=None):
...
-            self.input.ref_job = ref_job
+        self.input.from_hdf(self.project_hdf5)
+        self.input.ref_job = self[self.input.ref_job_name]
         self.output.from_hdf(self.project_hdf5)
 
     def interactive_close(self):
-        self[0].interactive_close()  # close the base job
-        self.to_hdf()   # run to_hdf to re-save input
-        self.output.to_hdf(self.project_hdf5)  # save output
+        self[self._base_name].interactive_close()  # close the base job
```
raynol-dsouza, line 290: This bit still complains that 'self[self._base_name]' is a 'ProjectHDFio' object.

raynol-dsouza:
I will merge this PR since it works with a serial protocol, and I can continue debugging why it complains while running parallely.

raynol-dsouza (approved):

PR #167, opened 2021-05-27
Add custom arrays to StructureContainer and use for TrainingContainer

pmrv:
Following the development of the flattened structure container in #148 I
1. moved the container to its own generic module
2. added the ability to add custom arrays to the `StructureContainer` that will be stored in the same flattened way
3. used it in the `TrainingContainer`

It turned out to be a bigger change, but I wanted to include the `TrainingContainer` changes directly to test the API.  hope I've kept the individual commits readable and nicely annotated, though.

There's two different types of arrays; per structure and per atom, that are resized independently.  You can allocate them with the `add_array` method or add together with the structure with the `add_structure` method and then access with `get_array()`. 

In the `TrainingContainer` I've kept the old storage in a pandas dataframe for now for convenience, but we could think about dropping that now, since it's not needed anymore.

@Leimeroth I think I've kept the container mostly compatible with your current usage in atomicrex, but please have a look if anything breaks.  The new custom arrays might also help you to store the extra fitting properties that you have in separate classes right now.

@yury-lysogorskiy I remember from the potential workshop that you also wanted to store neighbor list information in the container itself to avoid recomputing it between fitting runs.  This would be possible now in the `TrainingContainer`.  I thought about adding an optional argument to its `include_structure` method that accepts such a neighbor list as a per atom array, but we can also do it in another way if that would be more convenient.

The next step on my agenda would be to move the structure container to `pyiron_atomistic`.

This depends on [this](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations), so we should wait until the next atomistics bump until we merge this.


Leimeroth:
Atomicrex still works with these small fixes. I think my try except block should probably be replaced with some proper versioning.  Maybe a HasHDFStorage class similar to your HasStructure class would be helpful for that.

pmrv:
> Atomicrex still works with these small fixes. I think my try except block should probably be replaced with some proper versioning. Maybe a HasHDFStorage class similar to your HasStructure class would be helpful for that.

Interestingly that's sitting in my local repo to stew.  Won't make it in time for this, though.  I'll add a version check and then merge this.

The sparse arrays are a bit more complicated, so I will skip them for now, but it will stay on my radar.

tests/atomistics/job/test_structurecontainer.py
```diff
@@ -0,0 +1,110 @@
...
+from pyiron_atomistics import Project
+from pyiron_contrib.atomistics.atomistics.job.structurecontainer import StructureContainer
+import os.path
+import numpy as np
+
+class TestContainer(TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.project = Project(os.path.dirname(__file__))
```
liamhuber, line 11: If you're setting a project you should probably be inheriting from `TestWithProject` or `TestWithCleanProject` from `pyiron_atomistics._tests`

pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
```diff
@@ -172,9 +174,26 @@ def run_if_interactive(self):
 
     def to_hdf(self, hdf=None, group_name=None):
         super().to_hdf(hdf=hdf, group_name=group_name)
-        with catch_warnings():
-            self._table.to_hdf(self.project_hdf5.file_name, self.name + "/output/structure_table")
+        self._container.to_hdf(self.project_hdf5, "structures")
```
liamhuber, line 177: Yeah, this is beautiful. This is exactly what overrides of `to_hdf` should look like 😄

pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
```diff
@@ -172,9 +174,26 @@ def run_if_interactive(self):
...
 
     def from_hdf(self, hdf=None, group_name=None):
         super().from_hdf(hdf=hdf, group_name=group_name)
-        self._table = pd.read_hdf(self.project_hdf5.file_name, self.name + "/output/structure_table")
+        hdf_version = self.project_hdf5.get("HDF_VERSION", "0.1.0")
+        if hdf_version == "0.1.0":
+            table = pd.read_hdf(self.project_hdf5.file_name, self.name + "/output/structure_table")
+            self.include_dataset(table)
+        else:
+            self._container = StructureContainer()
```
liamhuber, line 186: I think this line is not needed? `__init__` should take care of it before `from_hdf` gets called.

tests/atomistics/job/test_structurecontainer.py
```diff
@@ -0,0 +1,110 @@
...
+        R = np.ones(len(self.structures[0]))
+        self.cont.add_structure(self.structures[0], self.structures[0].get_chemical_formula(),
+                                energy=E, forces=F, pressure=P, fnord=R[None, :])
+        self.assertEqual(self.cont.get_array("energy", self.cont.num_structures - 1), E,
+                         "Energy returned from get_array() does not match energy passed to add_structure")
+        self.assertTrue(np.allclose(self.cont.get_array("forces", self.cont.num_structures - 1), F),
+                        "Forces returned from get_array() does not match forces passed to add_structure")
+        self.assertTrue(np.allclose(self.cont.get_array("pressure", self.cont.num_structures - 1), P),
+                        "Pressure returned from get_array() does not match pressure passed to add_structure")
+        self.assertTrue("fnord" in self.cont._per_structure_arrays,
```
liamhuber, line 79: for QoL I think it's nice to `f'print the {thing} that was wrong'` in these more general assertions that wouldn't otherwise print it, e.g. `self.cont._per_structure_arrays.keys()` or w/e here.

liamhuber (commented):
Very cool! 👍  I threw a bunch of red ink on it, but the only thing I feel sort of seriously about is utilizing `HasStorage` to abstract away more of the HDF management. I don't absolutely insist on this, but I would appreciate some serious consideration here. Everything else is just little nits/QoL stuff.

pmrv:
Note to myself, there seems to be a bug when writing, then reading the container that doesn't restore all structures.

pyiron_contrib/atomistics/atomistics/job/structurecontainer.py
```diff
@@ -0,0 +1,376 @@
...
+            frame (int, str): selects structure to fetch, as in :method:`.get_structure()`
+
+        Returns:
+            :class:`numpy.ndarray`: requested array
+        """
+
+        if isinstance(frame, str):
+            frame = self._translate_frame(frame)
+        if name in self._per_atom_arrays:
+            I = self._per_structure_arrays["start_indices"][frame]
```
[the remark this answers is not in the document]
pmrv: Just me being lazy, changed it.

pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
pmrv: This is forces `_container` to be empty again, if you ever call `from_hdf` on a `TrainingContainer`.  It's an edge case for sure, though.

pyiron_contrib/atomistics/atomistics/job/structurecontainer.py
```diff
@@ -0,0 +1,376 @@
...
+
+from itertools import chain
+
+import numpy as np
+import h5py
+
+from pyiron_atomistics.atomistics.structure.atoms import Atoms
+from pyiron_atomistics.atomistics.structure.has_structure import HasStructure
+
+class StructureContainer(HasStructure):
```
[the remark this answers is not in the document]
pmrv: I like it generally, but there's currently a problem storing numpy arrays of unicode strings, hence my en/decoding gymnastics in `to_hdf`/`from_hdf`.  I haven't located yet whether this is in out HDF layer or h5io.  With this problem I'd have to overwrite `HasStorage` methods anyway to do the conversion.  As I said elsewhere I have `from_hdf`/`to_hdf` mixin on the back burner and would be touching this sooner or later again anyway, so I'd prefer delaying this a little bit, even though I agree `HasStorage` is the way to go.

pmrv:
One point about naming.  There's a `StructureContainer` already in pyrion_atomistic that is a proper job.  For backwards-compat we probably don't want to rename it, but I also don't want this class to become a job, but rather do it as in the `TrainingContainer` and make a job that contains this class.  So I'd have rename this class somehow, what about `StructureContainerCore`, `StructureStorage` or `ArrayStorage`?

liamhuber:
I like 'StructureStorage' best among those. And once it eventually inherits from HasStorage that will be a very sensible and intuitive name 😝😉

Leimeroth (approved):
From my side it is fine to merge this. At some point I will have to update my atomicrex class accordingly, but that can wait.

liamhuber (approved):

pmrv:
> From my side it is fine to merge this. At some point I will have to update my atomicrex class accordingly, but that can wait.

In another commit I renamed some of the property of the container, but I have updated the atomicrex code in structure_list.py accordingly.  If I've missed anything let me know and we can fix it in another PR.

PR #172, opened 2021-06-14
Add workflow to update the environment.yml in dependabot PRs

niklassiemer:


niklassiemer:
@muh-hassani  could you please make and add a new `DEPENDABOT_WORKFLOW_TOKEN`?

niklassiemer:
@muh-hassani could you also check if there is a `CODACY_PROJECT_TOKEN` since the Coverage workflow seems to expect but not find one? I am kind of confused since the coverage was working here, was it not?

muh-hassani:
> 
> 
> @muh-hassani could you also check if there is a `CODACY_PROJECT_TOKEN` since the Coverage workflow seems to expect but not find one? I am kind of confused since the coverage was working here, was it not?

I just checked and `CODACY_PROJECT_TOKEN` already exists in the list of the secrets. I have rerun the workflow. Let's see if the issue persists.

muh-hassani:
@niklassiemer it failed. I understood what the issue is. The branch you are trying to merge is from your own fork, where the token is not defined.

niklassiemer:
Thanks a lot! Sorry for not paying attention to where the branch actually is located...

PR #173, opened 2021-06-14
Interactive Structure Storage Master

pmrv:
Add a new job class like `StructureListMaster` from `pyiron_atomistics`, but using `StructureStorage` and an interactive job.

It's not quite done yet, but I wanted to discuss three issues:
1. the interactive reference job lives next to the master for now, because I couldn't figure out how to make it a child job.  @raynol-dsouza I imagine the protocols do similar setups all the time, how is it handled there?
2. There's no output on the master yet, how do we usually forward output from child jobs to master jobs.  Should I overload `master.output` to look into `master.ref_job.output` or do we copy the HDF5 content directly?
3. `VaspInteractive` doesn't seem to support `interactive_enforce_structure_reset`.  Is that because VASP can't deal with arbitrary structures during an interactive run (sphinx can) or is it not needed in this case? (Maybe @sudarsan-surendralal?)

### DIFF
--- a/pyiron_contrib/__init__.py
+++ b/pyiron_contrib/__init__.py
@@ -42,6 +42,7 @@ JOB_CLASS_DICT['Mlip'] = 'pyiron_contrib.atomistics.mlip.mlip'
 JOB_CLASS_DICT['LammpsMlip'] = 'pyiron_contrib.atomistics.mlip.lammps'
 JOB_CLASS_DICT['MlipJob'] = 'pyiron_contrib.atomistics.mlip.mlipjob'
 JOB_CLASS_DICT['Atomicrex'] = 'pyiron_contrib.atomistics.atomicrex.atomicrex_job'
+JOB_CLASS_DICT['StructureListMasterInteractive'] = 'pyiron_contrib.atomistics.atomistics.job.structurelistmasterinteractive'
 
 
 from ._version import get_versions
--- a/pyiron_contrib/__init__.py
+++ b/pyiron_contrib/__init__.py
@@ -42,7 +42,7 @@ JOB_CLASS_DICT['Mlip'] = 'pyiron_contrib.atomistics.mlip.mlip'
 JOB_CLASS_DICT['LammpsMlip'] = 'pyiron_contrib.atomistics.mlip.lammps'
 JOB_CLASS_DICT['MlipJob'] = 'pyiron_contrib.atomistics.mlip.mlipjob'
 JOB_CLASS_DICT['Atomicrex'] = 'pyiron_contrib.atomistics.atomicrex.atomicrex_job'
-JOB_CLASS_DICT['StructureListMasterInteractive'] = 'pyiron_contrib.atomistics.atomistics.job.structurelistmasterinteractive'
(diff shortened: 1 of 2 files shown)


raynol-dsouza:
@pmrv, to answer (1.)

It could work if you do not save the reference job itself to the hdf5 file; save the ref `job_name` to the container, remove the ref job from the container, call to_hdf(). And then after calling from_hdf(), assign the ref job back again from its loaded `job_name`.

For protocols however, the method above ran into issues. As a workaround, we save the reference job separately, and pass it's `path` to the protocol to load the reference job, and then create child jobs.

sudarsan-surendralal:
> `VaspInteractive` doesn't seem to support `interactive_enforce_structure_reset`. Is that because VASP can't deal with arbitrary structures during an interactive run (sphinx can) or is it not needed in this case? (Maybe @sudarsan-surendralal?)

I guess what this does is to be able to change the structure (incl. the number of atoms and species) interactively? If I remember correctly, in VASP you can only change the positions interactively. Am I right @jan-janssen?

pmrv:
Thanks for the quick comments!

> @pmrv, to answer (1.)
> 
> It could work if you do not save the reference job itself to the hdf5 file; save the ref `job_name` to the container, remove the ref job from the container, call to_hdf(). And then after calling from_hdf(), assign the ref job back again from its loaded `job_name`.
> 
> For protocols however, the method above ran into issues. As a workaround, we save the reference job separately, and pass it's `path` to the protocol to load the reference job, and then create child jobs.
> 
> EDIT: the first part was @liamhuber's idea. We tried it about two weeks ago, but unfortunately ran into issues.

If I understand this correctly though, that would just make sure the job is defined after HDF5.  It would still live in the same project as the master job (and not in a sub project), right?

What I want is more or less
```
Project
 \ Master Job
   \ Child Reference Job
```
whereas it's now
```
Project
 | - Master Job
 \ Child Reference Job
```
i.e. they are siblings in the hierarchy.

I know it can be done (`Murnaghan` does it), but I cannot reproduce that behavior.  (`Murnaghan` is a `ParallelMaster`, so it's a bit more involved to copy the code around).

> > `VaspInteractive` doesn't seem to support `interactive_enforce_structure_reset`. Is that because VASP can't deal with arbitrary structures during an interactive run (sphinx can) or is it not needed in this case? (Maybe @sudarsan-surendralal?)
> 
> I guess what this does is to be able to change the structure (incl. the number of atoms and species) interactively? If I remember correctly, in VASP you can only change the positions interactively. Am I right @jan-janssen?

Yeah, that's what it does.  I had feared that. :/

liamhuber:
> i.e. they are siblings in the hierarchy.
>
> I know it can be done (Murnaghan does it), but I cannot reproduce that behavior. (Murnaghan is a ParallelMaster, so it's a bit more involved to copy the code around).

Yeah, I've gotten consistent results -- especially subclassing off of `FlexibleMaster` -- having everything as siblings. Some time ago I fought a bit trying to get them set up as HDF children ala Murnhaghan, but I also never managed.

pmrv:
I've used this for a while now with the "children" as siblings and it worked fine so far.  If no one objects I'll merge this on Monday.

PR #174, opened 2021-06-14
Updated protocols

raynol-dsouza:
This PR contains:

-A single protocol for TILD (Parallel and Serial versions) instead of HarmonicTILD and VacancyTILD.
-Add DecoupledOscillators job.
-Update notebooks.
-Protocol loading fix.
-Add convergence criteria for NEB.
-Add dynamic gamma for NEB -- my modified version of backtracking line search.
-Add parallel NEB for use with VASP.
-Fix data storage explosion for reference Hessian, DecoupledOscillators in TILD.
-Move protocol notebooks to a new directory, to bypass shell notebooks execution (since we do not want to run tests for vasp+lammps jobs, and they are not available in the testing environment).


liamhuber (commented):
Can you make a little integration test using the variable gamma to relax something simple (e.g., a single displaced atom) and compare it to static gamma?

raynol-dsouza:
Also, the notebooks, which work perfectly fine on my PC and the cluster, do not execute in in our tests here. I haven't managed to get it fixed so far, so for the time being, until everything is working A-OK, I thought it best to leave this PR as a draft.

liamhuber:
> @liamhuber I think the protocol minimize example, which uses the GradientDescent vertex should be a good enough test, no? I will get that example updated.

Yep, should be. When I read it it looked like you need a new vertex to use this functionality (i.e. by pointing the `gamma` input of the descent vertex to the output of the gamma vertex), so it thus needs an update. It should be possible to also add this functionality right inside the gradient descent itself, but I don't know if that then requires more IO to the descent vertex. For now having it as a separate vertex is anyhow fine.

raynol-dsouza:
@liamhuber I managed to get a better fix working for saving/loading jobs, and the test notebooks run on my local system, the head node and the queue. Here are again the key areas of what this PR modifies:

-A single protocol for TILD (Parallel and Serial versions) instead of `HarmonicTILD` and `VacancyTILD`.
-Add `DecoupledOscillators` job.
-Update notebooks.
-Protocol saving/loading fix.
-Add convergence criteria for NEB.
-Add dynamic gamma for NEB -- my modified version of backtracking line search.
-Add parallel NEB for use with VASP.
-Fix data storage explosion for reference `HessianJob`, `DecoupledOscillators` in `TILD`.
-Move protocol notebooks to a new directory, to bypass shell notebooks execution (since we do not want to run tests for vasp+lammps jobs, and they are not available in the testing environment).

No hurry for the review. 

On my TODO list in other PRs are:
-a nice way to run some crucial tests for the protocols, all vertices, and saving/loading the associated hdf files.
-get scipy's optimizers integrated to give the NEB protocols a choice of minimizers.
-get non-equilibrium Frenkel-Ladd + Reversible Scaling set up as protocols, such that the method can be used for non-lammps jobs.

PR #194, opened 2021-07-05
Add coverage combine

pmrv:
Coverage is missing source annotations, hopefully this fixes it.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,3 +2,4 @@
 [run]
 source = pyiron_contrib
 omit = pyiron_contrib/_version.py
+concurrency = multiprocessing
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -38,6 +38,7 @@ jobs:
     - name: Coverage
       shell: bash -l {0}
       run: |
+        coverage combine
         coveralls
         coverage xml
         python-codacy-coverage -r coverage.xml


pmrv:
Don't seem to help.  Anyone have an idea why this doesn't work as in atomistics even though the config is the same? :S

jan-janssen:
To my understanding the combine is only required when parts of the calculation are executed as subprocesses - e.g. `non_modal` mode. This does not seem to be the case here.

pmrv:
Yeah, I was just trying things and that's the only difference I saw.  I've read [here](https://github.com/lemurheavy/coveralls-public/issues/572) that it might be a token issue.  How's the login work to coveralls or is there any reason why the token for this repo should have changed at all?

jan-janssen:
To me the output seems correct, or ?

pmrv:
There coverage summary seems to be correct, but if I go to the per line results for a single file ([here](https://coveralls.io/builds/41124278/source?filename=pyiron_contrib%2Fatomistics%2Fatomistics%2Fjob%2Fstructurestorage.py#L159) it says "no source found", where as in atomistics this works [e.g. here](https://coveralls.io/builds/41128536/source?filename=pyiron_atomistics%2Flammps%2Fbase.py#L789).

jan-janssen:
When I use the @pyiron-runner Account to login to coveralls.io everything is fine, while in the public version the source file is not visible. 
Screenshot 2021-07-08 at 09 23 09
I assume it is an issue on the coveralls site.

jan-janssen:
Also with my personal GitHub account I can access the file. Can you try to login to coveralls with your GitHub account?

pmrv:
> Also with my personal GitHub account I can access the file. Can you try to login to coveralls with your GitHub account?

Works after I login.  I didn't know that it works with the normal github account.  It's a bit strange that for contrib you have to login and for atomistics not, but naja.

PR #196, opened 2021-07-15
Use DataContainer for structure storage in RandSpg

pmrv:
This leaves the API compatible, but breaks HDF compatibility.  I know we have some RandSpg examples from workshop out in the wild, but supposedly these are pinned to specific version of contrib, so unless someone insists, I will not add backwards compatibility.


niklassiemer:
@pyiron/release On the current release I bump minor, since the `broken compatibility` tag is present here, probably we should have this tag in the other repos as well since it makes it very obvious when to bump minor.

PR #197, opened 2021-07-15
Force cells and pbcs to be saved per chunk and add test

pmrv:


pmrv:
Pending [this](https://github.com/pyiron/pyiron_base/pull/383).

jan-janssen:
To simplify the review process, I would suggest to mark pull requests with dependencies as draft. Still now as `pyiron_base` version 0.2.21  is merged, I guess the dependency issue is resolved for this pull request.

pmrv:
The issue here was that it was against a master before `FlattenedStorage` was moved to base.  I opened a new PR #218 against the current master with this commit cherry-picked, so closing.

PR #201, opened 2021-07-18
Compatibility to old atomicrex jobs

Leimeroth:
The or expression is not evaluated completely since an error is thrown when the key does not exist in the hdf file. I am not sure if anyone besides me uses this for production calculations, so I guess I can load all atomicrex jobs and use job.to_hdf using this branch instead of adding this into the actual code.
Alternatively I would suggest adding it to the code but additionally printing a deprecation warning and the workaround to make sure that anyone using this can apply it and than we can remove it when bumping version numbers at some point.
I am fine with both, just want to hear what you would prefer @pmrv

### DIFF
--- a/pyiron_contrib/atomistics/atomicrex/structure_list.py
+++ b/pyiron_contrib/atomistics/atomicrex/structure_list.py
@@ -196,8 +196,12 @@ class ARStructureContainer:
                 num_atoms = h["flattened_structures/num_atoms"]
                 group_name_2 = "flattened_structures"
             else:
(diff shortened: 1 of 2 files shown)


pmrv (commented):
Looks good to me.  I seemed to have thought that `hdf[key]` gives `None` for non-existing keys, but I should have checked that, so good catch!

I agree we want to drop this at some point, since this is only used in atomicrex, but as long as this is in contrib, I'm in no hurry.

Leimeroth:
Than I will merge it and we can take it out when fully integrating it into pyiron_base

PR #205, opened 2021-07-22
Add StorageJob to access data via S3

niklassiemer:


niklassiemer:
To receive files from the s3 the changes in https://github.com/pyiron/pyiron_base/pull/379 are needed. Then I can also derive `S3FileData` from `FileDataTemplate` of base.

niklassiemer:
@pmrv I tested with this branch and the current master of `pyiron_base` the following code 
```
from pyiron_base import Project
from pyiron_contrib.RDM.storagejob import StorageJob

with open('some_file.txt', 'w') as f:
    f.write('some text')

pr = Project('test')
job = pr.create.job.StorageJob('test_local')
job.add_files('some_file.txt', metadata={'awesome':"True"})

pr['test_local']
```
and got the error posted below. Probably, fist have a look into my implementation of `from`/`to_hdf`. Maybe I messed something up there.

```
ValueError                                Traceback (most recent call last)
<ipython-input-12-c3b132c9421d> in <module>
----> 1 pr['test_local']

/mnt/c/Users/Siemer/pyiron_git/pyiron_base/pyiron_base/project/generic.py in __getitem__(self, item)
   1383                         item=item_lst[0], convert_to_object=True
   1384                     ).__getitem__("/".join(item_lst[1:]))
-> 1385         return self._get_item_helper(item=item, convert_to_object=True)
   1386 
   1387     def __repr__(self):

/mnt/c/Users/Siemer/pyiron_git/pyiron_base/pyiron_base/project/generic.py in _get_item_helper(self, item, convert_to_object)
   1414             if self._inspect_mode or not convert_to_object:
   1415                 return self.inspect(item)
-> 1416             return self.load(item)
   1417         if item in self.list_files(extension="h5"):
   1418             file_name = posixpath.join(self.path, "{}.h5".format(item))

/mnt/c/Users/Siemer/pyiron_git/pyiron_base/pyiron_base/project/generic.py in load(self, job_specifier, convert_to_object)
    755             s.logger.warning("Job '%s' does not exist and cannot be loaded", job_specifier)
    756             return None
--> 757         return self.load_from_jobpath(
    758             job_id=job_id, convert_to_object=convert_to_object
    759         )

/mnt/c/Users/Siemer/pyiron_git/pyiron_base/pyiron_base/project/generic.py in load_from_jobpath(self, job_id, db_entry, convert_to_object)
    776         if job_id:
    777             job = jobpath(db=self.db, job_id=job_id, user=self.user)
--> 778             job = job.load_object(
    779                 convert_to_object=convert_to_object, project=job.project_hdf5.copy()
    780             )

/mnt/c/Users/Siemer/pyiron_git/pyiron_base/pyiron_base/job/core.py in load_object(self, convert_to_object, project)
    551             with project.open("..") as job_dir:
    552                 job_dir._mode = "a"
--> 553                 return self.to_object(project=job_dir, job_name=self._name)
    554         return self
    555 

/mnt/c/Users/Siemer/pyiron_git/pyiron_base/pyiron_base/job/core.py in to_object(self, object_type, **qwargs)
    486                 + "\" is empty, so it can not be loaded."
    487             )
--> 488         return self.project_hdf5.to_object(object_type, **qwargs)
    489 
    490     def get(self, name):

/mnt/c/Users/Siemer/pyiron_git/pyiron_base/pyiron_base/generic/hdfio.py in to_object(self, class_name, **qwargs)
   1259 
   1260         obj = self.create_instance(class_object, **qwargs)
-> 1261         obj.from_hdf(
   1262                 hdf = self.open(".."),
   1263                 group_name = self.h5_path.split('/')[-1]

/mnt/c/Users/Siemer/pyiron_git/pyiron_contrib/pyiron_contrib/RDM/storagejob.py in from_hdf(self, hdf, group_name)
    214     def from_hdf(self, hdf=None, group_name=None):
    215         super().from_hdf(hdf, group_name)
--> 216         self._stored_files.from_hdf(hdf=self._hdf5, group_name=group_name)
    217         self._input.from_hdf(hdf=self._hdf5, group_name=group_name)
    218         self._storage_type = StorageType(self._input.storage_type, self._input.storage_type_read_only)

/mnt/c/Users/Siemer/pyiron_git/pyiron_base/pyiron_base/generic/datacontainer.py in from_hdf(self, hdf, group_name)
    725         if version == "0.1.0":
    726             self.read_only = bool(hdf.get("read_only", False))
--> 727             self.update(hdf["data"], wrap=True)
    728         else:
    729             self.read_only = bool(hdf.get("READ_ONLY", False))

/mnt/c/Users/Siemer/pyiron_git/pyiron_base/pyiron_base/generic/hdfio.py in __getitem__(self, item)
    746                         obj = hdf_item.copy()
    747                         return obj
--> 748                 raise ValueError("Unknown item: {} {} {}".format(item, self.file_name, self.h5_path))
    749             else:
    750                 if item_lst[0] == "":  # item starting with '/', thus we have an absoute HDF5 path

ValueError: Unknown item: data /mnt/c/Users/Siemer/pyiron/projects/StorageJob/test/test_local.h5 /test_local/test_local
```

pyiron_contrib/RDM/storagejob.py
```diff
@@ -0,0 +1,272 @@
...
+            self._external_storage = FileS3IO(
+                config=self._input.s3.config,
+                path=self.path,
+                bucket_name=self._input.s3.bucket_name
+            )
+
+    def to_hdf(self, hdf=None, group_name=None):
+        super().to_hdf(hdf, group_name)
+        self._stored_files.to_hdf(hdf=self._hdf5, group_name=group_name)
+        self._input.to_hdf(hdf=self._hdf5, group_name=group_name)
```
pmrv, line 229: Generally `from_hdf`/to_hdf` expect to write/read themselves into/from `hdf[group_name]` and have that group exclusively for themselves.  In case of the `DataContainer` passing `group_name` overwrites `table_name` as the target HDF group.  Therefore here both `_input` and `_stored_files` try to write to the same group, I think.  That this does not raise an error is a bug.

I would usually write something closer to this
```python
# from here on self._hdf5 points to hdf[group_name]
self._input.to_hdf(hdf=self._hdf5) # writes to self._hdf5[self._input.table_name]
# OR!
self._input.to_hdf(hdf=hdf, group_name='my_input') # writes to self._hdf5['my_input']

**Comment by pmrv** [comment]
Not exactly sure what's going on, but since the `DataContainer` is trying to read from HDF with the backwards-compatible code, it suggests something is written inconsistently.  I made a comment above.

If I understand correctly `FileS3IO` and `ProjectS3IO` parallel `FileHDFio` and `ProjectHDFio`, right?  In that case we should map that in the code and have two common base classes for this, maybe `FileIO` and `ProjectIO`.  The nice thing about this would be that we then could pass any `ProjectIO` to an object's `to_hdf` (which is then incorrectly named, but should work).  It's ok though to get this in first and then extract common features in the base classes in a separate PR, I think.

**Comment by niklassiemer** [comment]
Thanks a lot for taking the time!

Yes these classes parallel the hdf ones. However, so far I do not have a proper use case for `ProjectS3IO`. Thus, I would drop it for this PR and introduce it later when needed.
The problem I see for the s3 is that you cannot do `s3[key]=value` since everything in s3 has to be file like...

File: pyiron_contrib/RDM/storagejob.py (hunk 1 above)
Comment (niklassiemer) [reply pyiron_contrib/RDM/storagejob.py:229]: Thanks a lot! I did not look up how the group_name is being used in the super call... I will drop the group names in the `DataContainer.to/from_hdf` calls and see if this works out 👍

**Comment by pmrv** [comment]
> Yes these classes parallel the hdf ones. However, so far I do not have a proper use case for `ProjectS3IO`. Thus, I would drop it for this PR and introduce it later when needed.

Yes, later is fine, but we should think about it already.

> The problem I see for the s3 is that you cannot do `s3[key]=value` since everything in s3 has to be file like...

I don't have a feeling what the best solution would be.  I suppose we'll find some way to turn everything we have into `bytes` (worst case: pickle/dill), but does it make sense to create a new value in S3 for each key-value pair or should be coalesce at some level.  It's not so urgent for now, I suppose.

**Comment by niklassiemer** [comment]
It should for sure be collected in some way. Each object on the s3 store has additional Metadata and, thus, quite some overhead in terms of storage. For our jobs it sounds reasonable to me to actually store everything in hdf fist and transfer the hdf if the job is finished.

**Comment by niklassiemer** [comment]
@pmrv I have some issue with the `read_only` of my `DataContainer` subclass... On `from_hdf` the mode is set to `read_only` first, before the actual data is read in (at least if I read [this](https://github.com/pyiron/pyiron_base/blob/064f1eb985fdaff49b1426266a7f1dc0aac0cd0c/pyiron_base/generic/datacontainer.py#L729) correctly). Since I want to fail hard on trying to modify the storage location of an in use `StorageJob`, this causes my overall `from_hdf` to fail all the time. So what did I do wrong here?

**Comment by pmrv** [comment]
> @pmrv I have some issue with the `read_only` of my `DataContainer` subclass... On `from_hdf` the mode is set to `read_only` first, before the actual data is read in (at least if I read [this](https://github.com/pyiron/pyiron_base/blob/064f1eb985fdaff49b1426266a7f1dc0aac0cd0c/pyiron_base/generic/datacontainer.py#L729) correctly). Since I want to fail hard on trying to modify the storage location of an in use `StorageJob`, this causes my overall `from_hdf` to fail all the time. So what did I do wrong here?

No, I think that's a bug in `DataContainer`, I don't think `read_only` sees a lot of usage.  Reading and setting the `read_only` flag simply after the contents should work a-ok and be simply a matter of exchanging the corresponding statements in the code you linked.

**Comment by niklassiemer** [comment]
Pending #215 as this branch skips simply skips the broken test and updates `pyiron_base` to access the newest features from there. Otherwise I am happy with the current status and encourage a review :)

File: pyiron_contrib/RDM/storagejob.py [hunk 2]
```diff
@@ -0,0 +1,324 @@
...
+
+        Parameters:
+            storage_type(str): one of ['local', 's3']
+        """
+        super().__init__(init=init, table_name=table_name)
+        self._available_storage_types = ['local', 's3']
+        self._storage_type = None
+        self.storage_type = storage_type
+
+    def create_group(self, name):
```
liamhuber, line 28: IMO this method should be pushed up to `DataContainer`. Your thoughts @pmrv?

pyiron_contrib/RDM/storagejob.py
```diff
@@ -0,0 +1,324 @@
...
+
+    @storage_type.setter
+    def storage_type(self, storage_type):
+        if storage_type not in self._available_storage_types:
+            raise ValueError(f"Expected 'storage_type' to be one of {self._available_storage_types} "
+                             f"but got {storage_type}.")
+        self._storage_type = storage_type
+        for storage in self._available_storage_types:
+            self[storage] = False
+        self[storage_type] = True
```
liamhuber, line 63: I think this is cool, but it makes me nervous. If I understand correctly 

```python
store = StorageType()
store.storage_type = 'S3'
print(store.S3)
>>> True
```

works, but 

```python
store = StorageType()
store.S3 = True
print(store.storage_type)
>>> None
```

...or? i.e. we allow two ways to read but only one way to set, with no clear warning that setting in the secondary way won't work!

You might be able to get around this by messing with the generic setters and screening for keys in `_available_storage_types`, but I'm not convinced it's worth it.

pyiron_contrib/RDM/storagejob.py
```diff
@@ -0,0 +1,324 @@
...
+
+
+class StorageJob(GenericJob):
+    """Job to store files associated with meta data either locally, or at a remote data service like s3."""
+
+    def __init__(self, project, job_name):
+
```
liamhuber, line 70: Nit: remove newline

pyiron_contrib/RDM/storagejob.py
```diff
@@ -0,0 +1,324 @@
...
+        self._storage_type = StorageType(storage_type='local', table_name="storage_type")
+        self._external_storage = None
+
+    def _before_generic_remove_child(self):
+        if self._storage_type.s3:
+            _remove_s3_working_directory(self._external_storage, self.path)
+
+    def check_setup(self):
+        """
+        this function is called in self._run_if_new() if self.server.run_mode.queue
```
liamhuber, line 83: Nit: sentence capitalization "This"

pyiron_contrib/RDM/storagejob.py
```diff
@@ -0,0 +1,324 @@
...
+        run_if_scheduler is the function called to sent the job to the cluster, however, it already sends the 
+        stuff to the queue which is obviously not intended if this is used as ssh storage system.
+        TODO: 
+        - overwrite run_if_scheduler to not trigger a job to be run on the queing system. 
+        - make some file_handler class to allow for 'store/receive/delete file' with a uniform interface for
+            - local
+            - s3
+            - ssh
+        """
+        raise NotImplementedError("Storing files remotely is not yet supported.")
```
liamhuber, line 96: I am A-OK with this for this PR, but this docstring really feels like it needs to be in our high-level developer docs somewhere...!!!

pyiron_contrib/RDM/storagejob.py
```diff
@@ -0,0 +1,324 @@
...
+            else:
+                self._logger.warning("Storage NOT empty - Danger of data loss!")
+        self._external_storage = external_storage
+        self._storage_type.storage_type = 's3'
+        self._storage_type.create_group("s3_config")
+        self._storage_type.s3_config.config = config
+        self._storage_type.s3_config.bucket_info = self._external_storage.bucket_info
+        self._storage_type.s3_config.bucket_name = bucket_name
+
+    def _list_groups(self):
```
liamhuber, line 116: Shouldn't this still return the `"storage_type"` table? I thought child data containers should appear as groups.

pyiron_contrib/RDM/storagejob.py
```diff
@@ -0,0 +1,324 @@
...
+        self.run()
+
+    def run_static(self):
+        raise NotImplementedError('There is no static run mode for a Storage job.')
+
+    def collect_output(self):
+        pass
+
+    def append(self, job):
+        pass
```
liamhuber, line 228: I would prefer if this raised an error. I guess it's a mandatory method from the parent class and appending jobs is meaningless here, but without an error I worry that people will pass a string and think they've just successfully appended a file to storage.

pyiron_contrib/RDM/storagejob.py
```diff
@@ -0,0 +1,324 @@
...
+    def interactive_flush(self, path="generic", include_last_step=True):
+        pass
+
+    def run_if_refresh(self):
+        pass
+
+    def _run_if_busy(self):
+        pass
+
+    def write_input(self):
```
liamhuber, line 249: Man, all these passes smell to high heaven 😭

pyiron_contrib/generic/s3io.py
```diff
@@ -0,0 +1,564 @@
...
+    def _ipython_display_(self):
+        result = f"File    {self.filename}   with meta data:\n"
+        result += "-------------------------------------------\n"
+        for key, value in self.metadata.items():
+            result += f"{key}:  {value}\n"
+        print(result)
+
+
+class S3ioConnect:
+    def __init__(self, credentials=None, bucket_name=None):
```
liamhuber, line 66: This is a bit hard to parse for the newcomer -- can you refactor some of this behaviour out into clearly named methods? I am imagining that they would be simple, only a two or three lines, but giving a name to the behaviour will help readability.

pyiron_contrib/generic/s3io.py
```diff
@@ -0,0 +1,564 @@
...
+        if self.is_dir(path):
+            path = self._to_abs_bucketpath(path)
+        else:
+            raise ValueError("No valid path specified!")
+        if path[-1] != '/':
+            path = path + '/'
+        path = path + data_obj.filename
+
+        data = data_obj.data()
+        if metadata is None:
```
liamhuber, line 419: I'm not a huge fan of the double-if implicit "if it's *still* `None`". How do you feel about this two-liner? (I tried a one-liner starting with `metadata if metadata is not None else...` and it runs fine, but I think the line length is too long.)

```suggestion
        if metadata is None:
            metadata = data_obj.metadata if data_obj.metadata is not None else {}
```

pyiron_contrib/generic/s3io.py
```diff
@@ -0,0 +1,564 @@
...
+    def _list_objects(self):
+        return self._bucket.objects.filter(Prefix=self._bucket_path)
+
+    def print_fileinfos(self):
+        """
+            Prints the filename, last modified date and size for all files in the current group,
+            recursively including sub groups.
+        """
+        for obj in self._bucket.objects.filter(Prefix=self._bucket_path):
+            print('/{} {} {} bytes'.format(obj.key, obj.last_modified, obj.size))
```
liamhuber, line 438: Nit: no f-string?

pyiron_contrib/generic/s3io.py
```diff
@@ -0,0 +1,564 @@
...
+    @staticmethod
+    def print_file_info(filelist):
+        """
+            Prints filename, last_modified, and size of each file in the provided list of file objects.
+
+            Args:
+                filelist (list): List containing objects from a bucket.
+        """
+        for obj in filelist:
+            print('/{} {} {} bytes'.format(obj.key, obj.last_modified, obj.size))
```
liamhuber, line 470: f-string nit

liamhuber (changes requested):

pyiron_contrib/RDM/storagejob.py
```diff
@@ -0,0 +1,324 @@
...
+
+
+def _remove_s3_working_directory(file_s3_io_handle, group_to_remove):
+    # Upon moving s3 to base, this should be moved to pyiron_base/job/util.py
+    if file_s3_io_handle.is_dir(group_to_remove):
+        file_s3_io_handle.remove_group(path=group_to_remove)
+
+
+class StorageType(DataContainer):
+    def __init__(self, *, init=None, storage_type=None, table_name=None):
```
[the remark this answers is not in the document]
niklassiemer: In this case I explicitly want to break the 'normal' `DataContainer` behavior:
usually when calling `DataContainer(something)` the `init` of `DataContainer` is set to `something` and the `DataContainer` is updated with these values. Here, I want to restrict to `kwargs` only.

pyiron_contrib/RDM/storagejob.py
niklassiemer: `create_group` actually lives already on `DataContainer`. However, I need to overwrite this since by default the new group is of the same class (`self.__class__()`) as the original container. Here, I want to have a plain `DataContainer` without my restrictions as a sub group.

pyiron_contrib/RDM/storagejob.py
niklassiemer: Yes, you are right! That is indeed dangerous...

pyiron_contrib/RDM/storagejob.py
niklassiemer: Hmm... The `StorageJob` stores its `storage_type` in a `DataContainer` subclass. However, the job itself does not have a `group`, does it? I could of course provide the `self._storage_type` here. However, I do not see any user benefit of getting the config storage back? The only thing I would like here (and this could actually be implemented) would be a new 'folder' to store files here, i.e. another `StorageJob` where the path is extended by the name of the group.

pyiron_contrib/generic/s3io.py
```diff
@@ -0,0 +1,564 @@
...
+            raise ValueError("No valid path specified!")
+        if path[-1] != '/':
+            path = path + '/'
+        path = path + data_obj.filename
+
+        data = data_obj.data()
+        if metadata is None:
+            metadata = data_obj.metadata
+        if metadata is None:
+            metadata = {}
```
niklassiemer, line 422: This has to be removed, accordingly
```suggestion
```

pyiron_contrib/RDM/storagejob.py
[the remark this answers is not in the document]
liamhuber: Behaviour sounds good to me. It would be nice to see that intent documented somewhere though, as it's not immediately obvious from the code. Maybe at the class-level docstring, to highlight difference to the parent class?

pyiron_contrib/RDM/storagejob.py
liamhuber: > `create_group` actually lives already on `DataContainer`. 

Ah, so it does! My apologies, I thought I checked for that and found it missing but I must have been looking at the wrong class...

pyiron_contrib/generic/s3io.py
liamhuber: Indeed! I just wasn't aware that we could do multiline suggestions now 😄 AFAIK, last time I was reviewing it was limited to one line on the removal side. I'll learn how to do this!

liamhuber (approved):
Super!

liamhuber:
Some of the codacy nits are legit (trailing whitespace, and fstring that can be a regular string), but some are the obnoxiously unsatisfiable docstring complaints. You could make a pass reducing the codacy errors, but don't worry about eliminating them.

pyiron_contrib/RDM/storagejob.py
```diff
@@ -0,0 +1,337 @@
...
+from pyiron_contrib.generic.s3io import FileS3IO
+
+
+def _remove_s3_working_directory(file_s3_io_handle, group_to_remove):
+    # Upon moving s3 to base, this should be moved to pyiron_base/job/util.py
+    if file_s3_io_handle.is_dir(group_to_remove):
+        file_s3_io_handle.remove_group(path=group_to_remove)
+
+
+class StorageType(DataContainer):
```
pmrv, line 16: I'm not convinced this needs to inherit from `DataContainer`.  This only stores a single string or am I missing more?  Since nothing else is stored in this I'd ask this to either manually implement `to_hdf`/`from_hdf` or use a `DataContainer` to store the `storage_type` as in the [example](https://pyiron-base.readthedocs.io/en/latest/apidoc/pyiron_base.generic.datacontainer.html?highlight=DataContainer).  This avoid the weird issues with `create_group` and using `*` in the `__init__`.

pyiron_contrib/RDM/storagejob.py
```diff
@@ -0,0 +1,337 @@
+import os
+import posixpath
+import shutil
+
+from pyiron_base import GenericJob, DataContainer
+from pyiron_base.generic.filedata import FileData
+from pyiron_contrib.generic.s3io import FileS3IO
+
+
+def _remove_s3_working_directory(file_s3_io_handle, group_to_remove):
```
pmrv, line 10: Does it make sense have this live on `file_s3_io_handle`, so that `_before_generic_remove_child` below can just call the method?

pyiron_contrib/RDM/storagejob.py
pmrv: Agree, with @niklassiemer this would be the place have 'folders' of uploaded files, if necessary.

pyiron_contrib/RDM/storagejob.py
```diff
@@ -0,0 +1,337 @@
...
+            print(f"removed {files_to_remove}")
+        self.run()
+
+    def run_static(self):
+        raise NotImplementedError('There is no static run mode for a Storage job.')
+
+    def collect_output(self):
+        pass
+
+    def append(self, job):
```
pmrv, line 240: Unrelated to this PR, but we should probably kick this out of `GenericJob`.

pyiron_contrib/RDM/storagejob.py
```diff
@@ -0,0 +1,337 @@
...
+
+        name_lst = item.split("/")
+        item_obj = name_lst[0]
+        if item_obj in self._list_ext_childs():
+            child = self._hdf5[self._name + "_hdf5/" + item_obj]
+            print("job get: ", self._name + "_jobs")
+            if len(name_lst) == 1:
+                return child
+            else:
+                return child["/".join(name_lst[1:])]
```
pmrv, line 327: This 'job' will never have children, so I'd think this can go.
```suggestion
```

pmrv: Actually, it will also never have sensible (i.e. user-facing) data in HDF either, right?

pyiron_contrib/generic/s3io.py
```diff
@@ -0,0 +1,567 @@
...
+
+            Args:
+                credentials (str/dict/boto3.resource/None):
+                        if str: path to a json configuration file with login credentials for the bucket.
+                        if dict: dictionary containing the login credentials.
+                        if boto3.resource: This resource is used as is, requires bucket_name to be specified.
+                        if None: use Credentials as obtained by boto3, requires bucket_name to be specified.
+                bucket_name(str/None): Name of the bucket, overwrites name given in the config.
+
+            The configuration needs to provide the following information:
```
pmrv, line 77:
```suggestion
            The credentials needs to provide the following information:
```
Was that a typo?  Otherwise I'm not sure where to pass additional keywords, since this method does not accept kwargs.

pyiron_contrib/generic/s3io.py
```diff
@@ -0,0 +1,567 @@
...
+
+        Returns:
+            list: list of file names.
+        """
+        nodes = []
+        for obj in self._bucket.objects.filter(Prefix=self._bucket_path, Delimiter='/'):
+            nodes.append(obj.key.split('/')[-1])
+        return nodes
+
+    def list_all(self):
```
pmrv, line 239: Can be derived from `HasGroups`.

pyiron_contrib/generic/s3io.py
```diff
@@ -0,0 +1,567 @@
...
+        return new
+
+    def close(self):
+        """Close current group and open previous group."""
+        if len(self.history) > 1:
+            del self.history[-1]
+        elif len(self.history) == 1:
+            self.history[0] = "/"
+        else:
+            print("Err: no history")
```
pmrv, line 325: Could be silent or raise an error?  I'm not sure what the HDF5 equivalent does, so I'm open.

pyiron_contrib/generic/s3io.py
```diff
@@ -0,0 +1,567 @@
...
+        if not isinstance(credentials, dict):
+            raise TypeError(f"credentials is not one of the supported types but {type(credentials)}.")
+        return credentials.copy()
+
+    @property
+    def endpoint_url(self):
+        return self.s3resource.meta.client.meta.endpoint_url
+
+
+class FileS3IO:
```
pmrv, line 136: Some of the docstrings are indented one level too much.

Short class docstring with the main methods `download`/`upload`/`put`/... would be nice.

pmrv (changes requested):

pyiron_contrib/RDM/storagejob.py
niklassiemer: So far not planned. Could just provide a quick access to the meta data only.

pyiron_contrib/RDM/storagejob.py
niklassiemer: I moved it into `_before_generic_remove_child`

pyiron_contrib/RDM/storagejob.py
niklassiemer: I used the `DataContainer` since I like the features it has like the `.key` access to the fields, the already implemented `read_only`, and the easy creation of storage groups and their auto-hdf features I use in the `use_s3_storage().  However, right now it is probably not needed to store this in a `DataContainer` subclass at all since it would be possible to store it in a `job._store = DataContainer()` directly and just use a clever setter for the `_available_storage_types`.

pyiron_contrib/generic/s3io.py
niklassiemer: It is silent, I also silenced here.

pmrv (approved):

PR #211, opened 2021-07-30
Use StructureStorage instead of DataContainer for structures in RandSpg

pmrv:
Somehow I forgot about `StructureStorage` in https://github.com/pyiron/pyiron_contrib/pull/196, but I think it makes a lot of sense to use it, so I changed it again.

### DIFF
--- a/pyiron_contrib/atomistics/randspg/randspg.py
+++ b/pyiron_contrib/atomistics/randspg/randspg.py
@@ -7,9 +7,10 @@ from __future__ import print_function
 import os
 import posixpath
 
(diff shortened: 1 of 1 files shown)


niklassiemer:
@pmrv Is there something holding this PR back? It would fit to the minor bump for the next release since it also has the `broken compatibility` tag.

pmrv:
I will give it another quick test later tonight, but generally it's tested and should work.

niklassiemer:
Thanks a lot, it everything is fine, merge it and I trigger a release to 0.1.0 tomorrow morning 👍

PR #218, opened 2021-08-12
Force cells and pbcs to be saved per chunk and add test

pmrv:
Replaces #197 because that was against an old master.


pmrv:
Fix arrives in [base](https://github.com/pyiron/pyiron_base/pull/390).

PR #227, opened 2021-08-24
Calculate and parse per energy structures for test/train sets

pmrv:
Simply modifies the shell script to run a few extra commands to calculate the energy for every structure in the training set and collect them to HDF afterwards.

@muh-hassani I think this needs the new version of `mlip` and I'm not sure if you/we on the cluster have the sources.  I can provide them and a shell script to build them.

### DIFF
--- a/pyiron_contrib/atomistics/mlip/mlip.py
+++ b/pyiron_contrib/atomistics/mlip/mlip.py
@@ -10,7 +10,7 @@ import os
 import pandas as pd
 import posixpath
 import scipy.constants
-from pyiron_base import Settings, GenericParameters, GenericJob, Executable
+from pyiron_base import Settings, GenericParameters, GenericJob, Executable, FlattenedStorage
 from pyiron_atomistics import ase_to_pyiron
 from pyiron_contrib.atomistics.mlip.cfgs import savecfgs, loadcfgs, Cfg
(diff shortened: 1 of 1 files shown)


pmrv:
@muh-hassani I'd prefer a locally built version from source.  I saw that right now we have a clone of the old version in `.../pyiron/dev/mlip-dev`.  Instead I'd want the new version living on a (non-public) gitlab.  I guess the best steps would be:
1. Check out the new version (somehow with my gitlab credentials) into the folder above
2. build it
3. update the resources to reflect the new build deps (I can do that via github).

muh-hassani:
> 
> 
> @muh-hassani I'd prefer a locally built version from source. I saw that right now we have a clone of the old version in `.../pyiron/dev/mlip-dev`. Instead I'd want the new version living on a (non-public) gitlab. I guess the best steps would be:
> 
>     1. Check out the new version (somehow with my gitlab credentials) into the folder above
> 
>     2. build it
> 
>     3. update the resources to reflect the new build deps (I can do that via github).

So, I assume you take care of the first two steps, and I will do the last. Please just ping me, when the resource repo is updated on github so I will update the local one on the cluster. Please correct me, if I misunderstood you.

pmrv:
Can I simply mess around in `/u/system/SLES12/soft/pyiron/dev/mlip-dev` or will this break something (since it is owned by you).

PR #229, opened 2021-08-25
Move StructureStorage to atomistics

pmrv:
Added a dummy subclass, so that reading old instances from HDF still works.

Depends on [this](https://github.com/pyiron/pyiron_atomistics/pull/327).

@Leimeroth If I've done everything correct, nothing for you should change and even loading old jobs which used `StructureStorage` should work, but double check once a new version of atomistics is out.


pmrv:
@Leimeroth Ready for review now.

PR #268, opened 2021-11-16
Set the pyiron config for CI workflows

liamhuber:
Puts contrib CI in line with atomistics etc. by setting the pyiron config for our github workflows. Became an issue for compatibility tests over in [base #519](https://github.com/pyiron/pyiron_base/pull/519).

### DIFF
--- a/.ci_support/pyironconfig.py
+++ b/.ci_support/pyironconfig.py
@@ -1,0 +1,19 @@
+import os
+
+
+def main():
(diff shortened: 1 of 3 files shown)


muh-hassani:
@liamhuber, I know this is merged, but I have a question. If the addition of `.pyiron` file is necessary for the CI workflows, does this mean that the default `configuration` in case of absence of `.pyiron` file, results in failure of the CI tests?

PR #277, opened 2021-11-30
Test python 3.10

jan-janssen:
Waiting for conda-forge/scikit-image-feedstock#81


jan-janssen:
The `moto` package seems to be the issue - it is `noarch` but some of its dependencies are not yet available for python 3.10

jan-janssen:
Maybe the issue was just fixed https://github.com/conda-forge/docker-py-feedstock/pull/58

PR #278, opened 2021-12-01
Add Abstract Mixin for Potential Fitting

pmrv:
Here's a sketch for the common interface I imagine among potential fitting codes.  This is not set in stone, but should give us something to work with.  I want to insist on using TrainingContainer/FlattenedStorage, because I think it will be most flexible to implement generic solutions, but if there's missing functionality we can definitely work it out.  The `plot` method is basically just for show, we should discuss what kind of plots are reasonable (if any).

Sample implementation for MTPs will follow later today.

### DIFF
--- a/pyiron_contrib/atomistics/ml/potentialfit.py
+++ b/pyiron_contrib/atomistics/ml/potentialfit.py
@@ -5,6 +5,7 @@ Abstract base class for fitting interactomic potentials.
 import abc
 
 import pandas as pd
+import matplotlib.pyplot as plt
 
 from pyiron_base import FlattenedStorage
 from pyiron_contrib.atomistics.atomistics.job.trainingcontainer import TrainingContainer
--- a/pyiron_contrib/atomistics/ml/potentialfit.py
+++ b/pyiron_contrib/atomistics/ml/potentialfit.py
@@ -39,7 +39,7 @@ class PotentialFit(abc.ABC):
         else:
             raise ValueError("Data can only be added before fitting is started!")
 
-    @abstract
+    @abc.abstractmethod
     def _get_training_data(self) -> FlattenedStorage:
         pass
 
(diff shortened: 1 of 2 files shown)


Leimeroth:
I think we should implement all the "standard" plots that are typically used in machine learning papers, like trained vs dft energy,  trained vs dft forces for each dimension etc. but as seperate functions not in one big function returning all plots at once.
Maybe it would also be nice to have some default analysis methods for the training data set itself, to make it easier to identify which kind of data is missing.
On the implementation side I mainly see the problem that the fitting codes have are too different with respect to the properties they can fit, the way weights are applied and generally the amount of things that can be applied to have any kind of generic implementation even when using the structure container. F.e. I use a Structure container that stores the structures itself, but then I have a separate FlattenedStorage  for each property I want to fit, simply because it is theoretically possible to fit 16 different properties using atomicrex and for each property you can apply like 10 different options.

pmrv:
> I think we should implement all the "standard" plots that are typically used in machine learning papers, like trained vs dft energy, trained vs dft forces for each dimension etc. but as seperate functions not in one big function returning all plots at once. 

Agree, I'll have some my MTPs, so I'll add it there first and then we can see what we can extract from there.

> Maybe it would also be nice to have some default analysis methods for the training data set itself, to make it easier to identify which kind of data is missing. 

I also have plotting functions for this and been thinking about adding them to the `TrainingContainer` anyways.

> On the implementation side I mainly see the problem that the fitting codes have are too different with respect to the properties they can fit, the way weights are applied and generally the amount of things that can be applied to have any kind of generic implementation even when using the structure container. F.e. I use a Structure container that stores the structures itself, but then I have a separate FlattenedStorage for each property I want to fit, simply because it is theoretically possible to fit 16 different properties using atomicrex and for each property you can apply like 10 different options.

I've been meaning to add efficient 'split' and 'merge' functions to `FlattenedStorage`, then you could take a generic `TrainingContainer` split it into properties, add the options you want and when someone calls `training_data`/`predicted_data` you just merge them back together.

Leimeroth (approved):

PR #274, opened 2021-11-26
Calculate neighbors on TrainingContainer

pmrv:

PR #285, opened 2021-12-09
Add simple plotting methods to TrainingContainer

pmrv:
As discussed in #278 here some plotting functions for the `TrainingContainer`. I have some more that are pending on #274 will add those in a separate PR once it is through.

### DIFF
--- a/pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
+++ b/pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
@@ -224,3 +224,64 @@ class TrainingContainer(GenericJob, HasStructure):
         else:
             self._container = StructureStorage()
             self._container.from_hdf(self.project_hdf5, "structures")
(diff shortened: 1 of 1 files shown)


Leimeroth:
I like the plotting ideas. Did you test the runtime when calculating symmetries for a container with many structures? I could imagine that it takes really long when you run it on many structures, especially if some of them are relatively large.

pmrv:
I've tested on my sets of ~1000 and ~13000 structures and it's ~5s and ~55s respectively, so it's bearable.  Of course `plot_space_groups` only makes sense on fairly ordered structures which will be smaller necessarily.

jan-janssen:
Is it possible to move all plotting functions to a subclass? Then we could do `container.plot.cell()`, `container.plot.spacegroups()` and `container.plot.energy_volume()`. The advantage is that is simplifies the auto completion on the root level.

PR #286, opened 2021-12-10
First draft of cp2k interface

jan-janssen:
Copied from https://github.com/jan-janssen/pyiron-cp2k


jan-janssen:
I updated to repository https://github.com/jan-janssen/pyiron-cp2k to use this interface - so we can use it for testing

PR #287, opened 2021-12-10
First draft of quantum espresso interface based on ASE

jan-janssen:
Copied from https://github.com/jan-janssen/pyiron-quantum-espresso


jan-janssen:
The example is updated to use pyiron_contrib https://github.com/jan-janssen/pyiron-quantum-espresso

jan-janssen:
Suggestion from Marvin https://github.com/Lonitch/qeAPI

jan-janssen:
Cython interface for quantum espresso https://gitlab.com/shaoxc/qepy/-/tree/master

niklassiemer:
The FFT grid can be specified in the `&SYSTEM` section by providing values for `nr1, nr2, nr3` (hard grid) or `nr1s, nr2s, nr3s` (smooth grid) in `pw.x` calculations - see 
https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm294

jan-janssen:
Maybe also take a look at https://github.com/aiidateam/qe-tools

liamhuber:
@jan-janssen the only thing I'm not a fan of here and in #294 is that they use outdated input classes (`GenericParameters`, while #286 uses `DataContainer` and I already merged it). If you think it's feasible for you to update both/either of these PRs before merging that would be preferable; if you don't have time with this I'm also comfortable to just merge them as-is -- it is contrib after all.

PR #291, opened 2021-12-13
Mlip Descriptor Job

pmrv:
New job class to calculate the MTP descriptors of a structure set.  This relies on a resource-file that is not yet merged into the cluster resources, though.


jan-janssen:
@pmrv While it would be great to have tests, maybe it makes sense to merge this pull request now as others might also be interested in using it.

PR #314, opened 2022-01-17
Additional plots for TrainingContainer

pmrv:


Leimeroth (changes requested):
If we agree on the changes in #316 I would change self,._train._container occurences to self._train and than merge this

pmrv:
> If we agree on the changes in #316 I would change self,._train._container occurences to self._train and than merge this

Yeah, I'll rebase this once #316 is through.

Leimeroth:
Rebased again to include changed tests.
Can we merge it @pmrv ?

Leimeroth (approved):

PR #315, opened 2022-01-18
First draft of a RuNNer interface based on ASE

lxknll:
Hey everyone,

this branch contains a first draft for a revision of the RunnerFit Job class.

## Description
This updated class serves can, in principle be used standalone to generate a high-dimensional neural network potential from scratch using RuNNer. It heavily relies on our (very new) ASE calculator for RuNNer.  **However, this calculator has not been added to the public ASE repository up to now and remains in my private fork of the project (working on a pull request).** I am happy to provide anyone interested in testing my changes with access to my gitlab.com repo.

The updated class has access to all relevant options of the Runner Fortran code.

## Motivation and Context
At the end of last year, @pmrv, @Leimeroth, @mrshrbld and I discussed a possible extension of the RunnerFit Job class in order to make it more general.

## Related Issue
I created this pull request as a point of discussion, please let me know if you preferred an issue instead :)

## How Has This Been Tested?
For testing purposes, I have been loading an existing input.data and input.nn file. Complete usage examples can be found in the class docstring.

## Further Information
I have also made changes to the runner shell scripts [located in the pyiron-resources repo](https://github.com/pyiron/pyiron-resources/blob/master/runner/bin/run_runner_1.2.sh). I will upload them as soon as possible.

Thank you for your reviews and comments!
Cheers


lxknll:
I found it most challenging to get a grasp of the different storage options in pyiron (DataContainers, Parameters class, StructureStorage vs. StructureContainer, etc.). There is most definitely room for improvement so I am thankful for your tips! :blush:

pyiron_contrib/atomistics/runner/job.py
```diff
@@ -1,235 +1,444 @@
...
+__copyright__ = 'Copyright 2021, Georg-August-Universität Göttingen - Behler '\
+                'Group'
+__version__ = '0.1'
+__maintainer__ = 'Alexander Knoll'
+__email__ = 'alexander.knoll@chemie.uni-goettingen.de'
+__status__ = 'development'
+__date__ = 'January 7, 2022'
+
+
+class PotentialFittingBase(GenericJob):
```
pmrv, line 50: See [this](https://github.com/pyiron/pyiron_contrib/pull/278) pending PR.

pmrv (changes requested):
Looks great, thank you!

Am generally very happy with the PR, but I've marked up some "un-pyironic" things that you can't know about.

Big issues are:
- [ ] think about #278 already, whether you need additional functionality there or whether some methods would be difficult to implement; once I merge that rebase this PR on top
- [ ] simplify the HDF handling, probably `TemplateJob` from `pyiron_base` is a good inspiration
- [ ] the ASE calculator being unavailable for now

I'd rather not merge code that relies on unpublished code (i.e. dead code).  What's the time frame you think this will land in ASE?  If this is mid-term or longer (i.e. pre-workshop or just shortly before) can we make a separate (public) repository for your calculator.

pmrv:
> I found it most challenging to get a grasp of the different storage options in pyiron (DataContainer vs. Parameters, StructureStorage vs. StructureContainer, etc.). I think I now made sensbile choices, but feel free to correct me. There is most definitely room for improvement so I am thankful for your tips! blush

Point taken. :') We are restructuring things a lot in base and it hasn't settled into a steady state yet, so there we are.

pmrv:
Another thing that came to my mind: Does it make sense to couple mode 1&2 and 3 so tightly?  I imagine most of the time we would run mode 1&2 together once and then 3 a bunch of times for application.  If so, I would think it is better suited in a separate job class (which could derive from a `RunnerBase` if you're worried about duplicated things too much).

pyiron_contrib/atomistics/runner/job.py
```diff
@@ -1,235 +1,444 @@
...
+        """Initialize the class.
+
+        Args:
+            project (FIXME): The project container in which the job is created.
+            job_name (str):  The label of the job (used for all directories).
+        """
+        super().__init__(project, job_name)
+
+
+class RunnerInput(DataContainer):
```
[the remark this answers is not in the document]
lxknll: I agree, see 17cdfe4

pyiron_contrib/atomistics/runner/job.py
```diff
@@ -1,235 +1,444 @@
...
 
 
-class RunnerFit(GenericJob):
+class RunnerStructureContainer(StructureStorage):
+    """Store chemical structures as a Runner training dataset."""
+
+    __version__ = '0.3.0'
+    __hdf_version__ = '0.3.0'
+
     def __init__(self, project, job_name):
```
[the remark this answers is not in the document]
lxknll: I removed this in a0af945 . However, it is obsolete now because `RunnerStructureContainer` inherits from `TrainingContainer` with 0cea750 and cb33245

pyiron_contrib/atomistics/runner/job.py
```diff
@@ -1,235 +1,444 @@
...
-        self._training_ids = []
+        """Initialize the class.
+
+        Args:
+            project (FIXME): The project container in which the job is created.
+            job_name (str):  The label of the job (used for all directories).
+        """
+        super(RunnerStructureContainer, self).__init__()
+
+    def append(self, structure):
```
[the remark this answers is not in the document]
lxknll: I opened a merge request [here](https://github.com/pyiron/pyiron_atomistics/pull/518). If accepted, all properties needed for RuNNer fits (energy, forces, atomic charges), **except the totalcharge**,  would be automatically stripped from an ASE Atoms object. The total charge is dealt with separately in the `RunnerFit` class, because it is not a standard property in ASE calculators.

pyiron_contrib/atomistics/runner/job.py
```diff
@@ -1,235 +1,444 @@
...
+        self.add_structure(
+            struct,
+            dft_energy=dft_energy,
+            dft_forces=dft_forces,
+            dft_charges=dft_charges,
+            totalcharge=totalcharge
+        )
+        return struct
+
+    def to_ase(self):
```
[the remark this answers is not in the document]
lxknll: Unfortunately, I do need this at the moment, because the ASE calculator for RuNNer requires the total charge as a calculator property to function correctly. This is only handled through the `RunnerSinglepointCalculator` class right now. However, I agree that is tedious and will kick off a discussion with the ASE people once we have tested the ASE calculator a little bit more within our group.

pyiron_contrib/atomistics/runner/job.py
```diff
@@ -1,235 +1,444 @@
...
+                                          project data.
+            group_name (str, optional):   Subcontainer name.
+        """
+        super().to_hdf(hdf=hdf, group_name=group_name)
+
+        self.input.to_hdf(hdf=self.project_hdf5)
+        self.structures.to_hdf(hdf=self.project_hdf5)
+        self.output.to_hdf(hdf=self.project_hdf5)
+
+    def from_hdf(self, hdf=None, group_name=None):
```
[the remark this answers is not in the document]
lxknll: This should be accomplished with d493232 .

pyiron_contrib/atomistics/runner/job.py
```diff
@@ -1,235 +1,444 @@
...
+                                          project data.
+            group_name (str, optional):   Subcontainer name.
+        """
+        super().from_hdf(hdf=hdf, group_name=group_name)
+
+        self.input.from_hdf(hdf=self.project_hdf5)
+        self.structures.from_hdf(hdf=self.project_hdf5)
+        self.output.from_hdf(hdf=self.project_hdf5)
+
+    def restart(self, *args, **kwargs):
```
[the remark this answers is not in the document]
lxknll: Here I need a little bit more help. As I understand it, I need to extract `*args, **kwargs` and hand them over to the base class with `new_ham = super(RunnerFit, self).restart(*args, **kwargs)`. Could I somehow get around that? Thank you :)

pyiron_contrib/atomistics/runner/job.py
```diff
@@ -1,235 +1,444 @@
...
 
-AngstromToBohr = 1.88972612
-ElectronVoltToHartree = 0.03674932218
+    def __init__(self, *args, **kwargs):
+        """Initialize the class."""
+        super(RunnerInput, self).__init__(*args, **kwargs)
 
 
-class RunnerFit(GenericJob):
+class RunnerStructureContainer(StructureStorage):
```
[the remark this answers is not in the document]
lxknll: Perfect, I have tried to do this in 0cea750 and cb33245 .

lxknll:
I have spent some more time on the calculator and (hopefully) implemented most of what we discussed. I replied to your remarks above and mentioned the corresponding commits, hopefully this makes it easy enough to get an overview of my changes. 

As a general point, I think the `PotentialFit` mixin (#278) is looking quite nice. At the moment, I do not see any problems implementing the `get/add_training_data`, `get/add_predicted_data`, and `get_lammps_potential` routines for the RunnerFit class. I do have some ideas for interesting plots, but this is probably a point for later discussions. 

 PS: `restart_with_mode` is a great idea and still on my todo list ;)

pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
```diff
@@ -54,6 +54,10 @@ def __init__(self, project, job_name):
         self._container.add_array("forces", shape=(3,), dtype=np.float64, per="element")
         # save stress in voigt notation
         self._container.add_array("stress", shape=(6,), dtype=np.float64, per="chunk")
+        # Store total (per-structure) and atomic charge information.
+        self._container.add_array('totalcharge', dtype=np.float64, per='chunk')
+        self._container.add_array("charges", dtype=np.float64, per="element")
```
pmrv, line 59: Best to add these only once needed, otherwise this will increase storage needed.

pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
```diff
@@ -93,6 +101,21 @@ def include_job(self, job, iteration_step=-1):
...
+
+        # Retrieve total charge.
+        tc = job.output.totalcharge
+        if tc is not None:
+            totalcharge = tc[iteration_step]
+        else:
+            totalcharge = None
+
+        # Retrieve atomic charges.
+        cc = job.output.charges
```
pmrv, line 113: I'm not sure right now, if you normal `AtomisticGenericJob.output` has these always defined, so you might need to make a `hasattr` check.

pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
```diff
@@ -145,13 +176,18 @@ def include_dataset(self, dataset):
...
+            - totalcharge(float): charge of the whole structure
             - forces (Nx3 array of float): per atom forces, where N is the number of atoms in the structure
+            - charges (Nx3 array of floats):
             - stress (6 array of float): per structure stress in voigt notation
         """
         self._table_cache = self._table.append(dataset, ignore_index=True)
         # in case given dataset has more columns than the necessary ones, swallow/ignore them in *_
-        for name, atoms, energy, forces, stress, *_ in dataset.itertuples(index=False):
-            self._container.add_structure(atoms, name, energy=energy, forces=forces, stress=stress)
+        for name, atoms, energy, totalcharge, forces, charges, stress, *_ in dataset.itertuples(index=False):
```
pmrv, line 186: Since we're now talking about columns that may or may not be defined in the dataframe passed in, it's probably better to switch over to `iterrows` and then check which columns are defined/not None.

pyiron_contrib/atomistics/runner/job.py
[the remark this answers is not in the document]
pmrv: Fair let's leave it in then for now.

pyiron_contrib/atomistics/runner/job.py
```diff
@@ -293,10 +225,34 @@ def publication(self):
...
+        return self.storage.output
+
     @property
     def structures(self):
         """Store a dataset consisting of many chemical structures."""
-        return self._structures
+        return self.storage.structures
 
     @structures.setter
     def structures(self, structures):
```
pmrv, line 258: Should document that this appends structures, since this will not be visible in code, i.e.
```python
job.structures = [structure1, structure2]
job.structures = [structure3, structure4]
```
(where `structures` are `Atoms`) 
will look like it only contains 3 & 4, but actually contains all of them.

pyiron_contrib/atomistics/runner/job.py
```diff
@@ -306,8 +262,17 @@ def structures(self, structures):
             structures (list): A list of ASE Atoms objects or Pyiron Atoms
                                objects which are to be stored.
         """
+
         for structure in structures:
-            self._structures.append(structure)
+            energy = structure.get_potential_energy()
```
pmrv, line 267: Is this always defined?

pyiron_contrib/atomistics/runner/job.py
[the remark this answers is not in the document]
pmrv: `GenericJob.restart` only takes `job_name` and `job_type`, so simply wrap these explicitly.

pmrv (approved):
Sorry for the slow response, but it looks pretty good to me now.  I've left a few more nits, but that should not be too much work.

lxknll:
Hey I just wanted to give a quick update on this. Since last we spoke a lot of new developments have been made on the underlying ASE calculator for RuNNer and a PR has been made to the upstream repository two weeks ago (no comments yet). Among the changes was the introduction of separate data classes for storing the results and symmetry functions more conveniently. Unfortunately, these classes do not know how to write themselves to HDF5 yet which broke some of the previous progress.

The way I see it there are two options (@pmrv please weigh in on this if you see a better way :blush: )
1. add `to_hdf` and `from_hdf` functions to the data classes in the ASE calculator's code base. This is rather easily done (probably :D). However, afaik this is strays far from usual applications of ASE calculators and might not be accepted upstream.
2. Extend the interface on the pyiron-side to catch the data classes from the ASE calculator and handle their I/O to/from HDF5 explicitly.

What do you think will be the best way? 
@mrshrbld pinging you, too, for your opinion :wink:

pmrv:
> Hey I just wanted to give a quick update on this. Since last we spoke a lot of new developments have been made on the underlying ASE calculator for RuNNer and a PR has been made to the upstream repository two weeks ago (no comments yet). Among the changes was the introduction of separate data classes for storing the results and symmetry functions more conveniently. Unfortunately, these classes do not know how to write themselves to HDF5 yet which broke some of the previous progress.
> 
> The way I see it there are two options (@pmrv please weigh in on this if you see a better way blush )
> 
>     1. add `to_hdf` and `from_hdf` functions to the data classes in the ASE calculator's code base. This is rather easily done (probably :D). However, afaik this is strays far from usual applications of ASE calculators and might not be accepted upstream.
> 
>     2. Extend the interface on the pyiron-side to catch the data classes from the ASE calculator and handle their I/O to/from HDF5 explicitly.
> 
> 
> What do you think will be the best way? @mrshrbld pinging you, too, for your opinion wink

I would lean towards option 2.  I would think that you don't want pyiron-specific code in ASE and it'll be difficult to update it if/when we change interfaces on the pyiron side.

pmrv:
#278 is merged now.

pyiron_contrib/atomistics/runner/storageclasses.py
```diff
@@ -0,0 +1,353 @@
...
+        Matters are complicated further because `FlattenedStorage` only accepts
+        data chunks (read here: structures) where all different arrays of
+        information have the same length. Unfortunately, one can have different
+        numbers of symmetry functions for each element and varying numbers of
+        elements for each structure in a training dataset.
+
+        The solution here is to `pad` the sfvalues of each group of elements in
+        one structure with rows filled with `np.NaN`. This way, all arrays for
+        one structure have length `number_of_atoms` and can be efficiently
+        stored and retrieved even for large  datasets.
```
pmrv, line 69: Not sure I follow.  Let's say I have a structure with (Al, Al, Li) atoms, then a NNP might have 4 descriptors for Al, but only 2 for Li and therefore you want to store something like
```
Element |           SF_Al |   SF_Li
-----------------------------------
Al      |   1   2   3   4 | nan nan
Al      |   2   1   4   3 | nan nan
Li      | nan nan nan nan |   1   2
```
Correct?

pyiron_contrib/atomistics/runner/utils.py
```diff
@@ -0,0 +1,125 @@
...
+                properties['totalcharge'] = structure.calc.get_property(
+                    'totalcharge'
+                )
+            else:
+                properties['totalcharge'] = np.sum(properties['charges'])
+
+        # StructureStorage needs a `spins` property, but not all ASE Atoms
+        # objects have that.
+        if not hasattr(structure, 'spins'):
+            structure.spins = None
```
pmrv, line 100: We only save and restore `spins` when available.  If that gave you and error, it's a bug we should fix in `StructureStorage`.

pyiron_contrib/atomistics/runner/job.py
```diff
@@ -85,98 +71,114 @@ class RunnerFit(GenericJob, HasStorage):
...
+    )
+
+    def __init__(self, project: Project, job_name: str) -> None:
         """Initialize the class.
 
-        Args:
-            project (Project): The project container where the job is created.
-            job_name (str):  The label of the job (used for all directories).
+        Parameters
+        ----------
```
pmrv, line 162:
```suggestion
        Args:
```

We generally go with google style in the rest of the code base.

pmrv (commented):
Great, thanks for your changes!

>     1. white_check_mark  _`energy` should not be a mandatory argument_ in `add_structure`/`include_structure`. One might want to create a potential for force prediction alone or for electrostatics (atomic charge prediction). In these cases, energies would not be needed.

I'm open for that, but would probably want to wait until after the workshop unless necessary for our demonstrator there, just to give us more time to make it work smoothly.  In that case we should make an issue "PotentialFit wishlist" that tracks this kind of stuff.

>     2. white_check_mark _Proposition for simplified `include_structure`/`add_structure` routines._ We discussed on Wednesday that `TrainingStorage` can already store arbitrary per-element or per-chunk arrays. Therefore, I think it is not necessary to pass some properties as keyword arguments (forces, stress, ...) while others are collected in the `arrays` dictionary that is subsequently unpacked. I removed those keyword arguments. In principle, this should not affect compatibility with other job types. For now I did not touch the `to_pandas` routine to avoid breaking compatibility.

We should check then though that each call provides a minimum of necessary information.  With just bare `**kwargs` I get away with adding structures without any fitting properties and the code will only fail during the actually fitting run.  What I'd imagine is that every `PotentialFit` sub class defines a list of necessary and optional properties.  That's why I hadn't allowed `**kwargs` so far even though it's obvious. 

>     3. white_check_mark _Added `to_dict` routine._ At the moment, one can obtain a tuple of all structures in storage by calling `to_list`. However, we cannot predict the different property arrays people might want to store in the future and the tuple does not return the name of each array. I added a `to_dict` routine which created a dictionary instead of a tuple.

I'd actually like to avoid the `to_list()`, because of these ordering problems, but also because you end up having to deal with nested lists again.  For now the best option would probably to use the columns from the `to_pandas()` return value for predictable ordering.  The other (future) option would be to be able to directly access the `TrainingStorage` from the `TrainingContainer`, because the storage already gives you all kinds of options to access the data.  I've been hesitant to allow that directly so far, because then people can get the storage and mess it up (by accident) leaving the container in an inconsistent state.  I'm ok having the `to_dict()` for convenience now, but would like to axe it again then as soon as we have better ways of accessing the data.

>     4. grey_question  In `atomistics.runner.utils` I added two routines `container_to_ase` and `ase_to_container` which translate `TrainingContainer` into a of ASE Atoms objects (or the other way around). This might generally be an interesting addition to the `TrainingContainer` class?

My hope would be that at some point this would be as easy as
```python
map(pyiron_to_ase, storage.iter_structures())
for s in ase_atoms:
  storage.add_structure(s)
```

We could add the `SinglePointCalculator` from the training properties automatically when requested by a flag.  Imo that's also wish list material.

>     5. grey_question  One could think about overriding `__add__` for `TrainingStorage` so that two storage objects can be easily combined. This would also facilitate `add_training_data` a lot. At the moment, one has to manually unpack all the structures from the `TrainingContainer._container` and iteratively add them to `TrainingStorage` by calling `add_structure` over and over again.

@Leimeroth Added this [here](https://github.com/pyiron/pyiron_base/pull/749).  I wouldn't like to override `__add__`, though.

> Obviously I am lacking your overview about pyiron and pyiron_contrib so please let me know if you approve of those ideas. Otherwise, I can revert them and adapt the RunnerJob.
> 
> white_check_mark Apart from that, I added `TrainingContainer` and `TrainingStorage` to the initialization in `atomistics.job.__init__`. It was very convenient for testing, but we can also remove it again before merging.
> ### Changes to `PotentialFit` Mixin
> 
>     1. white_check_mark `get_lammps_potential` should have an optional keyword argument `element: Optional[List[str]] = None`. For RuNNer/n2p2, we need to specify a mapping between the elements in the training dataset and the LAMMPS atom types as part of the pair_style (`emap` parameter). Without an `element` parameter, this mapping always has to point to all elements in the training dataset. However, this breaks down when the training dataset is binary (Li and Al) but the simulation is only supposed to run for e.g. pure Li. In that case, the mapping points to an element that does not actually exist in the LAMMPS structure file.

I'm not sure whether that's necessary.  E.g. our lammps potential database (EAMs and such) always define all elements inside the potential dataframe and it's not causing issues.  I would guess lammps ignores type mappings when they're not present in the structure, but I really don't know.

>     2. white_check_mark _Changed `plot` function axis units to from "ev" to "ev / atom". Because you divide by the number of atoms?

Yep, good catch.

>     3. grey_question  _`plot` function should be renamed_. Maybe something more specific like `plot_rmse_energy`?

The idea would be to have something like `job.plot.whatever` as for the `TrainingStorage` in the future, feel free to rename for now though.

>     4. grey_question I am happy to help with adding other plotting routines if you are interested :)

I guess once the notebooks for the workshop take shape, we should sit down again and check what plots are common and flashy enough.

lxknll:
> Great, thanks for your changes!

Thanks for your comments! :D

> > 1. _`energy` should not be a mandatory argument_ in `add_structure`/`include_structure`.

> I'm open for that, but would probably want to wait until after the workshop unless necessary for our demonstrator there, just to give us more time to make it work smoothly. In that case we should make an issue "PotentialFit wishlist" that tracks this kind of stuff.

Fair enough, I will revoke that change in `TrainingContainer` and `TrainingStorage` and create a separate issue to start a wishlist :grin: 

> > 2.  _Proposition for simplified `include_structure`/`add_structure` routines._

> We should check then though that each call provides a minimum of necessary information. With just bare `**kwargs` I get away with adding structures without any fitting properties and the code will only fail during the actually fitting run. What I'd imagine is that every `PotentialFit` sub class defines a list of necessary and optional properties. That's why I hadn't allowed `**kwargs` so far even though it's obvious.

I understand, better to be careful. Is that something you would still be interested in until the workshop or put it off until afterwards? If you'd rather wait, I can revoke this, too.

> > 3. _Added `to_dict` routine._ 

> I'd actually like to avoid the `to_list()` [...] I'm ok having the `to_dict()` for convenience now, but would like to axe it again then as soon as we have better ways of accessing the data.

Sure, I also agree that `to_dict` it is just a temporary workaround. 

> > In `atomistics.runner.utils` I added two routines `container_to_ase` and `ase_to_container` 

> My hope would be that at some point this would be as easy as [...] We could add the `SinglePointCalculator` from the training properties automatically when requested by a flag. Imo that's also wish list material.

Looks great, I will make a note in the wish list.

> > 5. One could think about overriding `__add__` for `TrainingStorage` so that two storage objects can be easily combined.

> @Leimeroth Added this [here](https://github.com/pyiron/pyiron_base/pull/749). I wouldn't like to override `__add__`, though.

Awesome, looking forward to having that in the code base! Just out of curiosity, why would you refrain from using `__add__`? :smiley: 

> > `get_lammps_potential` should have an optional keyword argument `element: Optional[List[str]] = None`. For RuNNer/n2p2, we need to specify a mapping between the elements in the training dataset and the LAMMPS atom types as part of the pair_style (`emap` parameter). 

> I'm not sure whether that's necessary. E.g. our lammps potential database (EAMs and such) always define all elements inside the potential dataframe and it's not causing issues. I would guess lammps ignores type mappings when they're not present in the structure, but I really don't know.

You are right that LAMMPS just ignores superfluous type mappings. Unfortunately, afaik n2p2 does not. As an example here would be a simple pair_style for the LiAl dataset:
```sh
pair_style nnp emap "1:Li,2:Al"
```
However, if one tries to predict a structure that only contains Al, pyiron will write a `structure.inp` file with only one atom type. As a result, n2p2 will think that all atoms in `structure.inp` are Li (because emap says that atom type 1 is Li) and predict a wrong/no energy. (@mrshrbld am I right?) 
There is the option to leave out `emap` completely. Then, the ordering of atom types in `structure.inp` would have to be the same as in RuNNer, i.e. ordered by periodic number. Unfortunately, pyiron did not do this for the input files I saw.

Therefore, I still think that an `element` keyword cannot be avoided. If you had an idea how to avoid the keyword that would be great, though! :blush:    

> > 3. _`plot` function should be renamed_.

> The idea would be to have something like `job.plot.whatever` as for the `TrainingStorage` in the future, feel free to rename for now though.

I see, that would be really nice! In that case the generic plot name is no problem.

> > 4. I am happy to help with adding other plotting routines if you are interested :)
> I guess once the notebooks for the workshop take shape, we should sit down again and check what plots are common and flashy enough.

At the moment, our notebook contains a working example of a RuNNer with the current interface but without commentary. We will probably combine this with the detailed explanations of last year's notebook and then upload our draft to the Gitlab repository. I will let you know once that is done.

TO DO

- [ ] Keep the mandatory `energy` keyword for now.
- [ ] Maybe: revoke proposition for **kwargs in `add_structure`/`include_structure`.
- [ ] Create an issue "PotentialFit wishlist".
- [ ] Discuss whether we need the `element` keyword in `get_lammps_potential`.
- [ ] Discuss common and flashy plots for `PotentialFit` class.

pyiron_contrib/atomistics/runner/job.py
lxknll: Sorry, I missed that in the pyiron_base `CONTRIBUTING.rst`. I will change that in a future commit.

pyiron_contrib/atomistics/runner/utils.py
lxknll: I think it is because of line 202 f. in [pyiron_atomistics structurestorage.py](https://github.com/pyiron/pyiron_atomistics/blob/master/pyiron_atomistics/atomistics/structure/structurestorage.py):

```python
        if structure.spins is not None:
            arrays["spins"] = structure.spins
```

Maybe this could be solved with a simple `hasattr` check in the if-statement?

pyiron_contrib/atomistics/runner/storageclasses.py
```diff
@@ -0,0 +1,353 @@
...
+    def _to_hdf(self, hdf: ProjectHDFio) -> None:
+        """Write `self` to HDF5 storage.
+
+        Parameters
+        ----------
+        hdf : ProjectHDFio
+            The HDF file where `self` will be stored.
+        """
+        for idx, prop in enumerate(self.runnerase_properties):
+            hdf[f'{prop}__index_{idx}'] = self.__dict__[prop]
```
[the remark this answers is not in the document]
lxknll: I see, will be removed in the next commit.

pyiron_contrib/__init__.py
```diff
@@ -36,7 +36,7 @@
 JOB_CLASS_DICT['TrainingContainer'] = 'pyiron_contrib.atomistics.atomistics.job.trainingcontainer'
 JOB_CLASS_DICT['RandomDisMaster'] = 'pyiron_contrib.atomistics.mlip.masters'
 JOB_CLASS_DICT['RandomMDMaster'] = 'pyiron_contrib.atomistics.mlip.masters'
-JOB_CLASS_DICT['RunnerFit'] = 'pyiron_contrib.atomistics.runner.job'
+JOB_CLASS_DICT['RunnerJob'] = 'pyiron_contrib.atomistics.runner.job'
```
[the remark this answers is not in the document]
lxknll: I see your point :D @mrshrbld what is your take on this?

pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
```diff
@@ -63,26 +67,33 @@ def include_job(self, job, iteration_step=-1):
...
         Args:
             job (:class:`.AtomisticGenericJob`): job to take structure from
-            iteration_step (int, optional): if job has multiple steps, this selects which to add
+            iteration_step (int, optional): if job has multiple steps, this
+            selects which to add
         """
         self._container.include_job(job, iteration_step)
 
     @deprecate("Use add_structure instead")
-    def include_structure(self, structure, energy, forces=None, stress=None, name=None):
```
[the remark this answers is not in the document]
lxknll: That makes perfect sense, I will make another proposition in the next commit.

pyiron_contrib/atomistics/runner/storageclasses.py
lxknll: Yeah exactly. From `RunnerStructureSymmetryFunctionValues` (I know, I hate the name, too) I get one array of symmetry functions for each element. However, the length of each array is not the number of atoms in the structure but the number of atoms _of that element_ in the structure. In other words, it does not contain the `NaN` values from your table yet.

In order to store the symmetry functions at the right atom in `FlattenedStorage` I therefore have to add rows to each array. That is why I am calling the `pad` and `unpad` routines from `utils.py`.

Is that okay?

pmrv: Hm, for that kind of flexibility I can only think of long format tables right now, which is not what we want in `FlattenedStorage`.  So fine with me to leave this as is, until we have an idea.

pyiron_contrib/atomistics/runner/utils.py
pmrv: Weird, will investigate next week.

pmrv:
> > We should check then though that each call provides a minimum of necessary information. With just bare `**kwargs` I get away with adding structures without any fitting properties and the code will only fail during the actually fitting run. What I'd imagine is that every `PotentialFit` sub class defines a list of necessary and optional properties. That's why I hadn't allowed `**kwargs` so far even though it's obvious.
> 
> I understand, better to be careful. Is that something you would still be interested in until the workshop or put it off until afterwards? If you'd rather wait, I can revoke this, too.

I guess checking whether there's at least one kwarg that is not `identifier` should be fine, we can flesh it out later then.

> > > 5. One could think about overriding `__add__` for `TrainingStorage` so that two storage objects can be easily combined.
> 
> > @Leimeroth Added this [here](https://github.com/pyiron/pyiron_base/pull/749). I wouldn't like to override `__add__`, though.
> 
> Awesome, looking forward to having that in the code base! Just out of curiosity, why would you refrain from using `__add__`? smiley

I find operator overloading mostly useful if you can do some sort of algebra on the objects. That's not really the case for us, because neither `-` nor `*` have useful definitions for structure containers.  So it would just be there to save us some typing, but then implementing `__add__`, `__iadd__` and `__radd__` correctly and type-stable (if `A`, `B` are subclasses of `FlattenedStorage`, what type will `A() + B()` have? What about `B() + A()`?) seems a bit much, when `cont1.extend(cont2)` is perfectly readable and unambiguous.
 
> > > `get_lammps_potential` should have an optional keyword argument `element: Optional[List[str]] = None`. For RuNNer/n2p2, we need to specify a mapping between the elements in the training dataset and the LAMMPS atom types as part of the pair_style (`emap` parameter).
> 
> > I'm not sure whether that's necessary. E.g. our lammps potential database (EAMs and such) always define all elements inside the potential dataframe and it's not causing issues. I would guess lammps ignores type mappings when they're not present in the structure, but I really don't know.
> 
> You are right that LAMMPS just ignores superfluous type mappings. Unfortunately, afaik n2p2 does not. As an example here would be a simple pair_style for the LiAl dataset:
> 
> ```shell
> pair_style nnp emap "1:Li,2:Al"
> ```
> 
> However, if one tries to predict a structure that only contains Al, pyiron will write a `structure.inp` file with only one atom type. As a result, n2p2 will think that all atoms in `structure.inp` are Li (because emap says that atom type 1 is Li) and predict a wrong/no energy. (@mrshrbld am I right?) There is the option to leave out `emap` completely. Then, the ordering of atom types in `structure.inp` would have to be the same as in RuNNer, i.e. ordered by periodic number. Unfortunately, pyiron did not do this for the input files I saw.

It sounds like this is something we should handle better in the lammps job.  `StructureContainer` already makes sure, that structures from the same storage have consistent type indices  I'm ok then with allowing to specify the elements for now so we have a quick solution.

PR #316, opened 2022-01-21
Atomicrex standard plots

Leimeroth:
This allows to use @pmrv TrainingPlots class with the atomicrex interface.
I changed the TrainingContainer and TrainingPlots class to use the StructureStorage directly instead of the TrainingContainer jobs ._container.
This allows me to use the functionality within atomicrex with 3 lines of code and I also think it makes more sense since all plotting functionality is related to the structures and not to job functions.
I am not sure why TraininContainer inherits from GenericJob, besides historical reasons. Is it only so that it is part of the job_table()?
IMO we could create a subclass TrainingStorage that directly defines the necessary arrays and methods that are added within the TrainingContainer job and then only define functionality that requires TrainingContainer to be a GenericJob within the job class to clearly separate them. Right now I think that is pretty much all the functionality besides the run functions and the hdf functionality.

EDIT:
moved functionality to TrainingStorage in #317 to make this more clear


pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
```diff
@@ -231,7 +231,7 @@ def plot(self):
         """
         :class:`.TrainingPlots`: plotting interface
         """
-        return TrainingPlots(self)
+        return TrainingPlots(self._container)
```
pmrv, line 234: This is nice!

pyiron_contrib/atomistics/atomicrex/structure_list.py
```diff
@@ -418,6 +420,19 @@ def _parse_final_properties(self, struct_lines):
...
                         final_forces = np.empty((len_struct, 3))
     
+    @property
+    def plot(self):
+        """
+        :class:`.TrainingPlots`: plotting interface
+        """
+        return TrainingPlots(self._structures)
+
+    def prepare_plotting(self, final_values=False):
```
pmrv, line 430: Does it make sense to run that automatically in the `plot` property if not done earlier by the user explicitly?  This way they wouldn't need to remember to call this one first.

pmrv (approved):

pyiron_contrib/atomistics/atomicrex/structure_list.py
Leimeroth: Probably. I just don't know how to treat the final_values then? Right now you can call it using final_values=True. My idea would be to explicitly call prepare_plotting when using final values and within the property when using target values,  but I think it would make sense to agree on some common output handling before that. If i remember correclty you store output in a seperate container and it would be nice to have the interface similar I think

pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
Leimeroth: Should I merge #317 into this one?

pyiron_contrib/atomistics/atomicrex/structure_list.py
pmrv: I guess one way to do it would to use `FlattenedStorage.split` to have two storages for target and final values separately, but I'm not sure how well that slots with the rest of you setup.

Leimeroth: I think that would mess up some things, but I am not sure. Also I believe at some point it would be helpful to have target and final values accessible by the plotting class to create plots that show both values/their errors etc.

A clean and flexible but code intensive (in the sense of many lines, not complicated) way would probably be to ask everything that uses plotting to give guaranteed return values.
F.e. whoever wants to use the plot class has to make sure that .cells returns the array of cells .energies returns energies. and so on.
If such guarantees are documented well I guess it would be possible for all of us to have the structure storages we use return correct values using properties, while keeping the way the structures are stored internally completely flexible.
Alternatively we could require the fitting codes to have a to_standard_container() function on their structures and then document how this standard container has to look.

pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
Leimeroth: I will merge it to have everything in one pr

pyiron_contrib/atomistics/atomicrex/structure_list.py
pmrv: As we discussed I'll work on this in a separate PR.

pmrv (approved):
I've moved `_table` attribute across as well, since this doesn't break public API and is only used to cache `to_pandas`.  Otherwise I think this can be merged.

Leimeroth:
I will merge it once tests pass

pmrv:
I forgot, can you still put in the `include_structure` -> `add_structure` deprecation + the same for `include_job`?

Leimeroth:
> I forgot, can you still put in the `include_structure` -> `add_structure` deprecation + the same for `include_job`?

Yes, also forgot it already

PR #357, opened 2022-03-18
add unit to angle

Leimeroth:
Unit for angles in cell plot. @pmrv should I add a transformation to degrees instead of radians? I think it is more intuitive if there is a 90° angle than 1.57something radians.


pmrv:
Yes, sounds like a good idea to add an extra parameter (can be true by default).

PR #368, opened 2022-03-29
Sqs

samwaseda:


jan-janssen:
How does your class compare to https://github.com/pyiron/pyiron_atomistics/blob/master/pyiron_atomistics/atomistics/job/sqs.py ?

jan-janssen:
> I talked about this one at the last ADIS meeting, but in short, this one is especially for structures containing defects.

Not everybody had a chance to participate in that meeting, so sometimes it is helpful to add a couple of lines to your pull request to describe what you aim to achieve with your changes.

jan-janssen:
That does not mean we should not merge it

PR #406, opened 2022-05-17
Atomicrex inherit potentialfit

Leimeroth:


pyiron_contrib/atomistics/atomicrex/base.py
```diff
@@ -259,6 +253,30 @@ def potential_as_pd_df(self):
...
+        thereof).
+
+        Returns:
+            DataFrame: contains potential information to be used with a Lammps job.
+        """
+        return self.potential_as_pd_df()
+
+    @property
+    def plot(self):
+        return self.structures.plot
```
pmrv, line 278: I think until we have settled on a few generic plots for potentials, this should not be overwritten or at least it should be doing the same as original `PotentialFit.plot`.

pyiron_contrib/atomistics/atomicrex/structure_list.py
```diff
@@ -517,6 +526,83 @@ def prepare_plotting(self, final_values=False):
...
+        storage = container._container
+        atomic_energy_storage = FlattenedARScalarProperty(
+            num_chunks=storage.num_chunks, num_elements=storage.num_elements
+        )
+        atomic_energy_storage.num_chunks = storage.num_chunks
+        atomic_energy_storage.num_elements = storage.num_elements
+        atomic_energy_storage._per_chunk_arrays["fit"][0 : storage.num_chunks] = True
+        atomic_energy_storage._per_chunk_arrays["tolerance"][
+            0 : storage.num_chunks
+        ] = 0.001
```
pmrv, line 540: I guess you are breaking API here, because you want to set the same data more efficiently?  In that case, we should probably move this use case directly to `FlattenedStorage`, maybe by letting `set_array` accept `slice`.

I'm ok with leaving this for the workshop though, if you make an issue to remind us in base.

pmrv (approved):
Two small nits, but looks otherwise good to me.

pyiron_contrib/atomistics/atomicrex/base.py
Leimeroth: I also changed the PotentialFit plot method, because it felt weird to have a .plot on TrainingStorage and on the PotentialFit job class doing 2 completely different things. One is the interface to lots of plotting functionality which I really like and the one on PotentialFit only plots energies, which is nice but quite limited.

IMO we should change plot in PotentialFit to act as an interface similar to the TrainingStorage

pyiron_contrib/atomistics/atomicrex/structure_list.py
Leimeroth: This is necessary because it overwrites the standard values. In the way I normally add structures this doesn't matter, but typically energies are weighted higher than forces so I wanted a more reasonable choice for these values.

Leimeroth:
@pmrv  I added a sketch for a PotentialFit plotting interface. I didn't get to test it yet, just wanted to make my comment above more clear with it. It is inspired by your Plotting class for StructureStorage

Leimeroth:
@lxknll could you have a look at the PotentialStandardPlots class? I used @pmrv plot function and added a similar function for forces, so if the energy plotting function worked this should be compatible I assume, but maybe I overlooked something.
Also if you have any wishes regarding plots this is probably a good place for them

pmrv (approved):
I had some naming nits, but looks good to me!

pyiron_contrib/atomistics/atomicrex/structure_list.py
Leimeroth: Now I get it :)

pyiron_contrib/atomistics/atomicrex/general_input.py
```diff
@@ -649,28 +649,35 @@ def _to_xml_element(self):
         return algo
 
 
-class SpaMinimizer:
+class SpaMinimizer(DataContainer):
```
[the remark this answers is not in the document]
Leimeroth: This actually is one of 2 minimizers that need an additional local minimizer

pyiron_contrib/atomistics/atomicrex/structure_list.py
```diff
@@ -299,6 +303,11 @@ def get_vector_property(self, prop, identifier, final=True):
         else:
             return self.fit_properties[prop]._per_element_arrays["target_val"][slc]
 
+    def _sync(self):
```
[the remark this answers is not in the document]
Leimeroth: This is done to prevent the dangerous situation that num_elements and num_chunks are different throughout the several Flattened storages that I use to store properties. In most of the code I just use the indices stored in self._structures directly
and never actually change the num_elements or num_chunks, but here  I need to sync them to be able to use the extend function on the fit property FlattenedStorages.

TBH I think my current implementation offers a lot of potentially dangerous ways to mess with the code due to the history of this class and this could help to reduce these risks. Probably I should do a complete rework, maybe by modifying the strings directly and storing everything in 1 FlattenedStorage, or by defining everything as properties (also open for other suggestions) but that is for another PR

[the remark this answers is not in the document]
pmrv: Yeah, it can stay that way, I just want to have a feeling of what is needed in the class.

I thought about the concept of linking multiple storages together so that their arrays stay in sync, but felt that having one big storage is simpler.  It should also be fairly simple to break the big container into smaller ones again with `.split` in a property somewhere.

PR #410, opened 2022-05-27
Training storage extend iter

Leimeroth:
function to iterate over structures and properties stored in the training container simultaneously

### DIFF
--- a/pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
+++ b/pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
@@ -690,7 +690,7 @@ class TrainingStorage(StructureStorage):
         Yields:
             :class:`pyiron_atomistics.atomistitcs.structure.atoms.Atoms`, arrays: every structure attached to the object and queried arrays
         """
-yield from zip(self._container.iter_structures(), *(self._container.get_array_ragged(a) for a in arrays))
+        yield from zip(self._container.iter_structures(), *(self._container.get_array_ragged(a) for a in arrays))
 
     @property
     def plot(self):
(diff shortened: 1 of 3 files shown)


pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
```diff
@@ -248,6 +254,20 @@ def plot(self):
...
+
+        Args:
+            wrap_atoms (bool): True if the atoms are to be wrapped back into the unit cell; passed to
+                               :meth:`.get_structure()`
+            *arrays (str): name of arrays that should be iterated over
+
+        Yields:
+            :class:`pyiron_atomistics.atomistitcs.structure.atoms.Atoms`, arrays: every structure attached to the object and queried arrays
+        """
+        return self._container.iter(*arrays, wrap_atoms=wrap_atoms)
```
pmrv, line 269:
```suggestion
        yield from self._container.iter(*arrays, wrap_atoms=wrap_atoms)
```

pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
```diff
@@ -652,9 +678,33 @@ def to_list(self, filter_function=None):
...
+            s = self._get_structure(frame=i, wrap_atoms=wrap_atoms)
+            vals = [s]
+            for a in arrays:
+                vals.append(
+                    self.get_array(
+                        name=a,
+                        frame=i,
+                    )
+                )
+            yield vals
```
pmrv, line 703:
```suggestion
yield from zip(self._container.iter_structures(), *(self._container.get_array_ragged(a) for a in arrays))
```
maybe it's a bit too convoluted though.

pmrv (changes requested):
I like the idea! In fact something similar is on my backlog.  I'd rather not have it on the actual storages for now though, because I don't want to back this future development into a corner yet.
Anyways it should go on the `StructureContainer` in that case, but right now let's leave it on the `TrainingContainer`.  This way it's no difference to users and we don't have to bother with backwards-compatibility on the `StructureContainer` later.

PR #417, opened 2022-06-01
Feature/pacemaker2022

yury-lysogorskiy:


srmnitc:
@pmrv would you be able to review this quickly as we can then merge and release for the workshop.

pmrv (changes requested):
`_get_training_data` and `_get_predicted_data` should be fixed, otherwise things are ok.  I'll change the input to `DataContainer` after this is merged and then we can make a new release.

srmnitc:
@yury-lysogorskiy could you please fix the issues as soon as possible. We can then release and get everything ready. thanks!

pmrv:
The example notebook blocks the CI, since it won't run without training data, so I've removed it for now.  It's a very nice notebook though, so we should find a different place for it.

I also realized that I do not have a possibility to test pacemaker jobs on our cluster yet, so I won't change the input to `DataContainer`.  It'll be not as nice, but will save us from unnecessary back and forth.

I've made issues #419 and #420 to remind me to pick it up after the workshop.

PR #429, opened 2022-06-10
add pot_to_df function for meam

Leimeroth:


jan-janssen (approved):
Looks good to me

Leimeroth:
On the pyiron side this should work, the problem is the atomicrex part in this case. Currently it produces multiple spline files instead of one that is usable by lammps. In principle this is not hard to implement, but the documentation of the meam/spline file format is basically not existent so one has to try to reconstruct the correct order in which functions appear in the file from the lammps source code, which I didn't do yet.

PR #443, opened 2022-06-29
atomicrex file names

Leimeroth:
check for / in identifiers


jan-janssen (approved):
Looks good to me - I guess we should merge it as is.

PR #465, opened 2022-07-27
Tutorial niklas

niklassiemer:
Calculate binding energy example by @liamhuber taken from the [developer tutorial pycharm into](https://github.com/pyiron/pyiron_developer_tutorial/blob/main/resources/pycharm_intro/after.py).


srmnitc (changes requested):
Move ´after.py´ into the code folders

liamhuber:
IMO needs:
- Class formulation
- Input: Run control vars, e.g. re-run existing jobs
- Input: Pressure control
- More clever: exploit sequential running to use a better guess of the cell volume
  - Actually, the bulk run should basically *always* be zero pressure and its volume used in future calculations
- Shooting that last one in the foot: Parallelize to run all four components independently?
- Demo: Looping over cell size to compare NVT and NPT convergence rates

PR #577, opened 2023-02-02
Graph-based pyiron workflows

liamhuber:
As discussed on the teams page.

An example of Joerg's syntax running.

- [x] IO channels
- [x] IO interface
- [x] Node
- [x] Workflow
- ~~Session~~ This can actually probably just stay `Project`
- [x] Content (actual nodes)
- ~~Serialization~~ This can be a separate PR
- [x] QoL stuff I thought of over the weekend
  - [x] Rename `name` and `engine`, as these may frequently conflict with channel names
  - [x] Rename `input` and `output` to `inputs` and `outputs` to avoid conflicting with the verbs
  - [x] Make convenient iterators for the different layers, e.g. `for connection in channel`; `for channel in inputs`; `for node in workflow`
    - For inputs and outputs, it might be better to just directly rebase these onto a `DotDict` so they inherit  `keys(),`values()` and `items()`.
  - [x] Some sort of verification routine, to display the status and connections of all the ports in a IO/node/workflow (this comes for free with a GUI, but some text-based thing that gives a dataframe or something should be easy enough.)
  - ~~A tool using the inspect module to help node designers quickly check for naming consistencies between input/pre/engine/post/output (the error messages are not _so_ opaque, but this could be faster and easier).~~ inspect is used to generate the channels to start with now
- ~~Allow callables for `types` that can facilitate some custom type checking from the user~~ Opted to skip this for now since serialization would require saving an `exec`ing more raw code.
- [x] Make the node function central and mandatory
- [x] Don't make people use `ChannelTemplate` at all (i.e. scrape types from annotations fo the node function)
- [x] Expect node functions to return a value or tuple of values, not a dict
- [x] Update `nodes` module so child classes are created by modifying `__init__` instead of overriding class attributes
- [x] Update tests
- [x] Update docstring
- [x] Refactor out ugliness that crept in during the refactor to the new paradigm

Here:
- All the low-level in-memory stuff, working pretty well
- Some really good documentation, some missing docs
- Tests throughout
- An example

Big things not here:
- A good treatment of job creation
  - Project is just really deeply embedded in pyiron, understandably, and I'm not happy with the janky way I "avoid" it
- ~~QoL feedback, especially visualization~~ Let's add some quality -- TODOs updated
  - Debugging the example, I realized just how badly I miss the gui elements of ironflow!
- Serialization
  - This is absolutely critical, but IMO it should be a different PR. Either we can merge this and then work on that, or leave this un-merged and just stack the PRs.
- Macros
  - After thinking about it today, right now I lean towards a `to_node()` converter on the workflows, which dynamically generates you all the code needed to make a node with your current workflow's functionality. The big advantage is that it locks in what IO is open and what IO is internally connected within the workflow, which is the sort of robustness I'd like to see. I also think it may wind up really helping with "sharability", as then you have a light-weight python code snippet that other people can register with and use in their workflows. But I didn't want to bother implementing anything until we'd had some discussion.
- Discontinuous graph exectution, e.g. nodes that need to be explicitly "run" or execution inputs
- Any handling of using remote resources or starting subprocesses

### DIFF
--- a/notebooks/workflow_example.ipynb
+++ b/notebooks/workflow_example.ipynb
@@ -9,7 +9,7 @@
     {
      "data": {
       "application/vnd.jupyter.widget-view+json": {
-       "model_id": "409b55703383472687bef0084f85bc05",
+       "model_id": "0662258fcd7947dd8cc676649c899ac5",
        "version_major": 2,
        "version_minor": 0
       },
@@ -309,7 +309,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "2 2 2\n",
+      "1 1 1\n",
       "11 11 11\n"
      ]
     }
@@ -352,18 +352,18 @@
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "/Users/huber/work/pyiron/pyiron_contrib/pyiron_contrib/workflow/workflow.py:153: UserWarning: Reassigning the node whatever_n4_gets_used to the label n4 when adding it to the workflow my_wf.\n",
+      "/Users/huber/work/pyiron/pyiron_contrib/pyiron_contrib/workflow/workflow.py:166: UserWarning: Reassigning the node whatever_n4_gets_used to the label n4 when adding it to the workflow my_wf.\n",
       "  warn(\n"
      ]
     },
     {
      "data": {
       "text/plain": [
-       "{'n1': <pyiron_contrib.workflow.node.Node at 0x15208cb10>,\n",
-       " 'n2': <pyiron_contrib.workflow.node.Node at 0x1520823d0>,\n",
-       " 'n3': <pyiron_contrib.workflow.node.Node at 0x152092a90>,\n",
-       " 'n4': <pyiron_contrib.workflow.node.Node at 0x1520932d0>,\n",
-       " 'n5': <pyiron_contrib.workflow.node.Node at 0x152074f10>}"
+       "{'n1': <pyiron_contrib.workflow.node.Node at 0x1473cc050>,\n",
+       " 'n2': <pyiron_contrib.workflow.node.Node at 0x147131550>,\n",
+       " 'n3': <pyiron_contrib.workflow.node.Node at 0x1473ce090>,\n",
+       " 'n4': <pyiron_contrib.workflow.node.Node at 0x1473ce810>,\n",
+       " 'n5': <pyiron_contrib.workflow.node.Node at 0x1473cf050>}"
       ]
      },
      "execution_count": 12,
@@ -421,13 +421,9 @@
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "/Users/huber/work/pyiron/pyiron_contrib/pyiron_contrib/workflow/workflow.py:153: UserWarning: Reassigning the node bulk_structure to the label structure when adding it to the workflow with_prebuilt.\n",
+      "/Users/huber/work/pyiron/pyiron_contrib/pyiron_contrib/workflow/workflow.py:166: UserWarning: Reassigning the node bulk_structure to the label structure when adding it to the workflow with_prebuilt.\n",
       "  warn(\n",
-      "/Users/huber/work/pyiron/pyiron_contrib/pyiron_contrib/workflow/workflow.py:153: UserWarning: Reassigning the node lammps to the label engine when adding it to the workflow with_prebuilt.\n",
-      "  warn(\n",
-      "/Users/huber/work/pyiron/pyiron_contrib/pyiron_contrib/workflow/workflow.py:153: UserWarning: Reassigning the node calc_md to the label calc when adding it to the workflow with_prebuilt.\n",
-      "  warn(\n",
-      "/Users/huber/work/pyiron/pyiron_contrib/pyiron_contrib/workflow/workflow.py:153: UserWarning: Reassigning the node scatter to the label plot when adding it to the workflow with_prebuilt.\n",
+      "/Users/huber/work/pyiron/pyiron_contrib/pyiron_contrib/workflow/workflow.py:166: UserWarning: Reassigning the node lammps to the label engine when adding it to the workflow with_prebuilt.\n",
       "  warn(\n"
      ]
     },
@@ -436,6 +432,16 @@
      "output_type": "stream",
      "text": [
       "The job JUSTAJOBNAME was saved and received the ID: 298\n"
+     ]
+    },
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/Users/huber/work/pyiron/pyiron_contrib/pyiron_contrib/workflow/workflow.py:166: UserWarning: Reassigning the node calc_md to the label calc when adding it to the workflow with_prebuilt.\n",
+      "  warn(\n",
+      "/Users/huber/work/pyiron/pyiron_contrib/pyiron_contrib/workflow/workflow.py:166: UserWarning: Reassigning the node scatter to the label plot when adding it to the workflow with_prebuilt.\n",
+      "  warn(\n"
      ]
     },
     {
@@ -456,8 +462,7 @@
     "wf.plot = nodes.Scatter(\n",
     "    x=wf.calc.outputs.steps, \n",
     "    y=wf.calc.outputs.temperature\n",
-    ")\n",
-    "wf.update()"
+    ")"
    ]
   },
   {
--- a/notebooks/workflow_example.ipynb
+++ b/notebooks/workflow_example.ipynb
@@ -9,7 +9,7 @@
     {
      "data": {
       "application/vnd.jupyter.widget-view+json": {
-       "model_id": "2cf85a81cbaf46f4a0ce55cd4fb29f5f",
+       "model_id": "409b55703383472687bef0084f85bc05",
        "version_major": 2,
        "version_minor": 0
       },
@@ -352,18 +352,18 @@
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "/Users/huber/work/pyiron/pyiron_contrib/pyiron_contrib/workflow/workflow.py:137: UserWarning: Reassigning the node whatever_n4_gets_used to the label n4 when adding it to the workflow my_wf.\n",
+      "/Users/huber/work/pyiron/pyiron_contrib/pyiron_contrib/workflow/workflow.py:153: UserWarning: Reassigning the node whatever_n4_gets_used to the label n4 when adding it to the workflow my_wf.\n",
       "  warn(\n"
      ]
     },
     {
      "data": {
       "text/plain": [
-       "{'n1': <pyiron_contrib.workflow.node.Node at 0x10ae7b950>,\n",
-       " 'n2': <pyiron_contrib.workflow.node.Node at 0x14a29ae90>,\n",
-       " 'n3': <pyiron_contrib.workflow.node.Node at 0x14a29b590>,\n",
-       " 'n4': <pyiron_contrib.workflow.node.Node at 0x14a29bdd0>,\n",
-       " 'n5': <pyiron_contrib.workflow.node.Node at 0x14a290a10>}"
+       "{'n1': <pyiron_contrib.workflow.node.Node at 0x15208cb10>,\n",
+       " 'n2': <pyiron_contrib.workflow.node.Node at 0x1520823d0>,\n",
+       " 'n3': <pyiron_contrib.workflow.node.Node at 0x152092a90>,\n",
+       " 'n4': <pyiron_contrib.workflow.node.Node at 0x1520932d0>,\n",
+       " 'n5': <pyiron_contrib.workflow.node.Node at 0x152074f10>}"
       ]
      },
      "execution_count": 12,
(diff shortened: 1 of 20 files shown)


liamhuber:
[Binder link](https://mybinder.org/v2/gh/pyiron/pyiron_contrib/graph?labpath=notebooks%2Fworkflow_example.ipynb)

samwaseda:
I would say the content of `workflow.ipynb` should go to the DocString of `workflow.py` (because this DocString is currently quite abstract)

samwaseda:
Do you see a possibility of offering a function instead of `types`? For example, I have a function that takes either one temperature value or an array of temperature values, so that a function like this would make sense:

```python
def check_value(value):
    try:
        return np.min(value) >= 0
    except:
        return False
```

Ideally, I would love to overwrite `def ready`, but that doesn't seem to be straightforward.

liamhuber:
> Do you see a possibility of offering a function instead of `types`? For example, I have a function that takes either one temperature value or an array of temperature values, so that a function like this would make sense:
> 
> ```python
> def check_value(value):
>     try:
>         return np.min(value) >= 0
>     except:
>         return False
> ```
> 
> Ideally, I would love to overwrite `def ready`, but that doesn't seem to be straightforward.

@samwaseda, that is a great innovation. It should be super easy to add an `if` clause to `ready` (and `_types_to_tuple`) that checks if `types` are actually types or a callable and responds accordingly. 

I wouldn't want to directly override ready, but maybe we could ghost in a new method like `functional_type_checking` that child nodes can define in lieu of static types.

The only pain point I see, is that if users are passing in callables at instantiation time, the only option for (de)serialization is going to include another `exec` call. But we can cross that bridge and remove the functionality if needed when we get around to adding serialization.

samwaseda:
> The only pain point I see, is that if users are passing in callables at instantiation time, the only option for (de)serialization is going to include another `exec` call. But we can cross that bridge and remove the functionality if needed when we get around to adding serialization.

ah ok that's correct. Then I'm gonna for now include it in `preprocessor` and see how far I'm gonna come

pyiron_contrib/workflow/node.py
```diff
@@ -0,0 +1,252 @@
...
+        ...     ]
+        ...
+        ...     def __init__(self, name: str, engine: callable, **kwargs):
+        ...         # We'll modify what's available in init to push our users a certain direction.
+        ...         super().__init__(name=name, engine=engine, **kwargs)
+        >>>
+        >>> def add(x, y, z):
+        ...     return {"w": x + y + z}
+        >>>
+        >>> adder = ThreeToOne("add", add)
```
samwaseda, line 147: I would say the hurdle should be even lower for new users (i.e. if I have `def add`, I don't want to write the full class in addition). Here's my suggestion: We ask the user to write the function in the format:

```python
def add(x:int | float = 1, y:int | float = 2, z:int | float = 3) -> int | float:
    return x + y + z
```

Then we should be able to do

```python
adder = Node("add", add)
```

And internally you can do something like this:

```python
import inspect

new_node = Node()
for key, value in inspect.signature(add).parameters.items():
    new_input = {'name': key}
    if value.annotation is not inspect.Parameter.empty:
        new_input['types'] = value.annotation
    if value.default is not inspect.Parameter.empty:
        new_input['default'] = value.default
    new_node.input_channels.append(ChannelTemplate(**new_input))
```

Here, I didn't write the output channel, because I was not really sure how to cope with the fact that there's no name, but at least the type can be obtained by `inspect.signature(add).return_annotation`.

samwaseda: Note: The last cell should be rewritten when it's included inside `class Node`. I wrote it in this format so that it works in the notebook.

pyiron_contrib/workflow/channels.py
```diff
@@ -0,0 +1,140 @@
...
+    ):
+        self.name = name
+        self.node = node
+        self.default = default
+        self.value = default
+        self.types = None if types is None else self._types_to_tuple(types)
+        self.storage_priority = storage_priority
+        self.connections = []
+
+    @staticmethod
```
samwaseda, line 58: In my opinion this should not be needed, because if we use `typing` it takes care of whether it's one type or multiple types, i.e. the user should write either one type e.g. `int` or multiple types following `typing`, e.g. `int | float`

pyiron_contrib/workflow/node.py
samwaseda: @liamhuber ?

liamhuber: For input I like it, but it is problematic for output. Output is a dict, so even the type doesn't help us much. We might be able to nest types in the dict like like lost or tulle (I haven't tested this), but we're still in trouble because the dictionary is not ordered.

Ultimately I care more about the syntax for input and output being the same than for the syntactic sugar of not needing to specify channels, so if we can't figure a way to get the output running with this, I wouldn't really want it for input alone.

samwaseda: Why does output have to be a dict?

samwaseda: I have to say it makes a huge difference whether the user has to write this

```python
from pyiron_contrib.workflow.node import Node
from pyiron_contrib.workflow.channels import ChannelTemplate

class ThreeToOne(Node):
    input_channels = [
        ChannelTemplate("x", default=1, types=(int, float)),
        ChannelTemplate("y", default=2, types=(int, float)),
        ChannelTemplate("z", default=3, types=(int, float)),
    ]
    output_channels = [
        ChannelTemplate("w", types=(int, float)),
    ]
    
def add(x, y, z):
    return {"w": x + y + z}

adder = ThreeToOne("add", add)
```

or this

```python
from pyiron_contrib.workflow.node import Node

def add(x:int | float = 1, y:int | float = 2, z:int | float = 3) -> int | float:
    return x + y + z

adder = Node("add", add)
```

It's just not realistic to force people to write the upper variant.

liamhuber: You are right, it doesn't _have_ to be, but it's for basically the same rationale that the input is kwargs. If you are using just args as input and returning a tuple as output, then at the graph level you're writing connections like this `downstream_node.inputs[2] = upstream_node.outputs[2]`. 

This is actually how ryvencore is set up, and I can tell you from personal experience that it's a huge pain. Without names you lose track of what is going on so quickly. One of the very early things we did in ironflow was overload the ryven functionality to add dot-based access using the channel (port) names, and hit was a huge QoL improvement.

Similarly, if you think about a GUI representation of the graph, on the output side it just gets these boxes `output0; output1; output2`, so you have no immediate idea where they should go. You might show the output type, but there will absolutely be times where a function returns two types and you wouldn't know which is which without more carefully inspecting the innards of the node function (e.g. think about returning energies and volumes -- they'd both just be `np.array[float]`).

Actually, the more we talk about it the more convinced I am that forcing node developers to explicitly declare their channels with names is the right way to go. It's a bit more verbose, but rather than being a "hurdle" I think it's a help by making everything clear and explicit.

I took a quick look through the inspect docs, and I don't see any way to pull out the keys of a returned dictionary systematically. We could get the string source code and manually parse the string for stuff after `return`, but this seems _super_ fragile, e.g. if they define the dict early and then just `return my_dict`. Being robust would require a bunch of crazy code analysis.

liamhuber: > It's just not realistic to force people to write the upper variant.

I disagree. Also, `inspect` for sure won't handle anything with ontological typing, e.g. `ChannelTemplate('my_typed_thing', types=Atoms, otype=onto.AtomicStructure).

IMO there's nothing _hard_ or _complicated_ about the top example, it's just a bit verbose. People still write in C, so obviously there's plenty of folks willing to put up with some verbosity.

If you can find a way to get the bottom example working without throwing away the naming of channels at the graph level (i.e. output), I'm open to it. Although we'll need to be careful that the complexity we add by including all the implicit magic is clear enough to developers.

liamhuber: >  I don't want to write the full class in addition

Also you don't have to. Here is your lower example sans class. Definitely still more verbose, but keeps the output channel name.

```
from pyiron_contrib.workflow.node import Node
from pyiron_contrib.workflow.channels import ChannelTemplate
# In the long run we'll obviously give a single point of import for everything

def add(x:int | float = 1, y:int | float = 2, z:int | float = 3) -> dict:
    return {"w": x + y + z}

adder = Node(
    "add", 
    input_channels = [
        ChannelTemplate("x", types=(int, float)),
        ChannelTemplate("y", types=(int, float)),
        ChannelTemplate("z", types=(int, float)),
    ],
    engine=add,
    output_channels [ChannelTemplate("w", types=(int, float))]
)
```

samwaseda: > Also, `inspect` for sure won't handle anything with ontological typing, e.g. `ChannelTemplate('my_typed_thing', types=Atoms, otype=onto.AtomicStructure).

This argument is not valid, because not only we don't have it and also not in the example, but also even when we have it, we will most likely not make it mandatory. So the fact that the function definition cannot have the ontological typing does not make it less valuable. Besides there's definitely a way to define it as well:

```python
factory = Node("factory", create_atoms)
factory.my_typed_thing.otype = onto.AtomicStructure
```

> IMO there's nothing _hard_ or _complicated_ about the top example, it's just a bit verbose. People still write in C, so obviously there's plenty of folks willing to put up with some verbosity.

I disagree. In contrast to classes like `AtomicGenericJob`, which just have to be useful to developers, the workflow class must be easy to use for anyone making workflows, i.e. virtually everyone, if not literally everyone.

After having written this last phrase, I started realising that maybe this is where the difference between you and me comes from. For me, this new workflow class is on the one hand going to be used by developers to replace mainly the current master classes (in which case, yes I agree that they **must** explicitly say what are the output values), but also by anyone writing a code in the notebook, e.g. master students, early phd students or experimentalists who don't really have a clear idea of how to write a proper code, or also for advanced users who want to have a lot of small nodes quickly. In this case, it is important that a simple function can be included in the workflow in terms of pyiron.

samwaseda: Maybe we can also involve @pmrv and @srmnitc because they probably also have good insight into this.

liamhuber: > This argument is not valid, because not only we don't have it and also not in the example, but also even when we have it, we will most likely not make it mandatory. 

You're 100% write, I was being too hasty and am wrong. We [_do_ have it](https://github.com/pyiron/ironflow/blob/1e5195cdc68d80bd9256a002363baf8724addb72/ironflow/model/port.py#L56), but not having it in the "function" approach doesn't preclude having it in the OOP approach, so you'll get no further argument from me.

> After having written this last phrase, I started realising that maybe this is where the difference between you and me comes from. For me, this new workflow class is on the one hand going to be used by developers to replace mainly the current master classes (in which case, yes I agree that they must explicitly say what are the output values), but also by anyone writing a code in the notebook, e.g. master students, early phd students or experimentalists who don't really have a clear idea of how to write a proper code, or also for advanced users who want to have a lot of small nodes quickly. In this case, it is important that a simple function can be included in the workflow in terms of pyiron.

I don't think we have conflicting visions about who will use it, but maybe pretty different perspectives on how they should be treated. The pythonic experiment of treating everyone like an adult is a couple decades old, and has led to python being the language of choice for scientists and casuals, but IMO it's also a key contributor to so much scientific code being a hot pile of **** -- not even just at the architecture level, but even at the level of individual functions. It is precisely people "who don't really have a clear idea of how to write a proper code" who I have in mind when I make a framework where, at a minimum, the input and output of each piece of code need to be _named_. The tricky bit is that for a workflow paradigm, it's important to have named outputs as well as named inputs. That's perfectly in line with OOP, where you can populate class attributes, but the only way to accomplished named output at the function level in python is by returning a dict. 

Your example was shorter precisely because it dropped this piece of information. I agree that we don't want to force everyone to make subclasses, but we can recover it without forcing people to subclass `Node` as shown in my example above. I agree that it's a little bit verbose, but I don't think instantiating a second class (`ChannelTemplate`) is particularly difficult, and because of my perspective from the previous paragraph I am willing to pay some verbosity if in return we can force people to at least take a moment to name things.

That said, I don't have any objection in principal to your example of getting it all done in a single function call _if_ we can preserve (a) multiple output values, and (b) those output fields being explicitly named, and (c) the ability to associate types with the output fields (although I could be flexible on the last point). Ideally we could write something like 

```python
from pyiron_contrib.workflow.node import Node

def add(x:int | float = 1, y:int | float = 2, z:int | float = 3) -> dict["w": int | float]:
    return x + y + z

adder = Node("add", add)
```

But python just doesn't support this sort of behaviour, and I don't currently see any implementation that gets the job done.

liamhuber: > But python just doesn't support this sort of behaviour, and I don't currently see any implementation that gets the job done.

Actually, it looks like we can do it by type hinting [a TypedDict](https://peps.python.org/pep-0589/), but as far as I can see that involves first defining a TypedDict -- like, at the notebook level for your example, which I think kind of defeats the purpose and is just more verbose than defining a set of `output_channels`.

samwaseda: Look, this is a [Dask example they give on their website](https://examples.dask.org/delayed.html):

```python
def inc(x):
    return x + 1

def dec(x):
    return x - 1

def add(x, y):
    return x + y
```

And then they can be executed like this:

```python
x = inc(1)
y = dec(2)
z = add(x, y)
z
```

The first cell of this example, with the current implementation looks like this:

```python
def inc(x):
    return {'inc': x + 1}

def dec(x):
    return {'dec': x - 1}

def add(x, y):
    return {'add': x + y}


from pyiron_contrib.workflow.node import Node
from pyiron_contrib.workflow.channels import ChannelTemplate

class Inc(Node):
    input_channels = [
        ChannelTemplate("x"),
    ]
    output_channels = [
        ChannelTemplate("inc"),
    ]

inc_node = Inc('inc': inc)


class Dec(Node):
    input_channels = [
        ChannelTemplate("x"),
    ]
    output_channels = [
        ChannelTemplate("dec"),
    ]

dec_node = Dec('dec': dec)


class Add(Node):
    input_channels = [
        ChannelTemplate("x"),
    ]
    output_channels = [
        ChannelTemplate("add"),
    ]

add_node = Add('add': add)
```

And with my suggestion, it looks like this:

```python
def inc(x):
    return x + 1

def dec(x):
    return x - 1

def add(x, y):
    return x + y

from pyiron_contrib.workflow.node import Node

inc_node = Node('inc', inc)
dec_node = Node('dec', dec)
add_node = Node('add', add)
```

For someone who's used to working with Dask, my solution is probably acceptable, but I'm really not sure if they would be willing to do all the effort to write the channels, especially in this case they are just repetitions of names. A bigger problem here is that a usual workflow contains **a lot** of nodes, so this example is certainly not an exaggeration. Moreover, a function can be included into a workflow only if it's included in a node, so the user **must** do all the efforts, if my solution is not available, even if the function is only about adding two variables. Here, I'm not saying that pyiron workflow should be the same as the Dask workflow, but I have difficulties explaining why the user **must** ("must" and not "has the freedom to choose to") write so many more lines in pyiron.

liamhuber: All of those functions return a single value. Is there a solution in this format for multiple outputs?

liamhuber: Also, in your last example you don't show how to set up the graph. Is it supposed to look like this

```python
add_node.input.x = inc_node.output
```

or 


```python
add_node.x = inc_node
```

Again, I really run up against the fact that the output field has no name. I don't see that being workable for nodes with multiple outputs, and I don't think I want to abandon multiple outputs.

liamhuber: In a dask format, I suppose we could write something like 

```python
a, b = returns_multiple("some", "input")
c = another_function(a, b)
```

Which sort of handles the case for text-based interface -- but what about GUIs? We still just have "output0", "output1", etc...

liamhuber: I would finally argue that your "current implemenation" example still unfairly bloats lines by going the sub-class route. You can do everything just instantiating a `Node`, and wind up looking a lot closer to your final example:

```python
def inc(x):
    return x + 1

def dec(x):
    return x - 1

def add(x, y):
    return x + y

from pyiron_contrib.workflow.node import Node
from pyiron_contrib.workflow.channels import ChannelTemplate

inc_node = Node(
    'inc', inc, 
    input_channels=[ChannelTemplate('x')], 
    output_channels=[ChannelTemplate('y')]
)
dec_node = Node(
    'dec', dec, 
    input_channels=[ChannelTemplate('x')], 
    output_channels=[ChannelTemplate('y')]
)
add_node = Node(
    'add', add, 
    input_channels=[ChannelTemplate('x'), ChannelTemplate('y')], 
    output_channels=[ChannelTemplate('z')]
)

add_node.input.x = inc_node.output.y
add_node.input.y = inc_node.output.z
```

Again, I'm not saying this is perfect or that I'm closed to changes, but I absolutely want to preserve: named outputs (necessary for a GUI), and multiple outputs (necessary, e.g. for looping functions that need both value output and execution signal output). So far the much more succinct alternatives you're suggesting keep losing these things.

samwaseda: > Which sort of handles the case for text-based interface -- but what about GUIs? We still just have "output0", "output1", etc...

But that's up to the user to say whether it's good or not. I don't understand why we should not offer this possibility in the first place.

> The example above is 30 lines (including newlines to keep the instantiation arguments nicely readable), and your dask example is 14. So basically the current implementation pays a factor of 2 in verbosity to gain the features I want. That's a bit of a steep price, but it's not an order of magnitude or anything. Maybe we can get that factor down without losing functionality.

Wait my Dask example is 4 lines, including the import statement. Yours is 18 or 19, depending on the import. And I would say that's a lot more lines.

liamhuber: > But that's up to the user to say whether it's good or not. 

Me. I am the user. I have been working with this sort of workflow stuff for the last couple of years in various formats, and I'm telling you that refercing outputs by index (rather than by name) is a huge pain in the ass. I insist that outputs be named.

> I don't understand why we should not offer this possibility in the first place.

I'm not sure what you mean here.

> Wait my Dask example is 4 lines, including the import statement. Yours is 18 or 19, depending on the import. And I would say that's a lot more lines.

Yep, that's probably a better way to count. x5 instead of x2, then.

But it's still a bit beside the point, because the dask example is still missing functionality I want!

Here is a solution I'm not happy with, because I think it is not sufficiently explicit, relies on overloading, and does too much under-the-hood magic, but I could maybe live with it:

```python
def inc(x):
    return x + 1

def dec(x):
    return x - 1

def add(x, y):
    return x + y, x-y

from pyiron_contrib.workflow.node import Node
from pyiron_contrib.workflow.channels import ChannelTemplate

inc_node = Node(function=inc, output_channels=['y'])
dec_node = Node(function=dec, output_channels=['y'])
add_node = Node(
    fuction=add, 
    output_channels=['sum', 'diff'],
    x=inc_node.output.y,
    y=inc_node.output.y
)
```

Where we leverage `inspect` to populate the node name from the function name, and `input_channels` from the function signature. Further we change things so that the node `function` is expected to return a tuple (or single value), and the _order of `output_channels` matters_ and is used to map `function` output to `Node.output` _and_ we allow passing either a full `ChannelTemplate` _or_ just a string (which dynamically creates a `ChannelTemplate` under the hood (again, using `inspect` to populate the `types` if `function` had return type hinting).

We give up some of my "wants" of explicitness and directness and trade them in for dirty evil implicit misdirection, but we do keep all of my "needs" of mandatorily named output, multiple outputs, and giving power users the option to specify much more (e.g. ontological types). In return we get within a 2x of dask's line count -- 8 lines to dask's 4.

I expect nodes to be _used_ much more than they're _written_, and I like our `Node(..., arg=other_node.output.arg)` syntax a lot better than dask's `x = my_delayed_guy(y)`, and this solution also preserves our user syntax nicely.

liamhuber: >dirty evil implicit misdirection

(I'm being facetious, btw. It's true I have some reservations, but overall your arguments are pushing this in a better direction and I appreciate your persistence)

liamhuber: In fact, we could actually make construction of the channels the default, such that we _always_ construct the channels from the function signature, getting rid of the `types` and `input_channels` arguments altogether. The signature of  the node would then be something like

```python
Node.__init__(
    function: callable,  # Node functionality, returns value or tuple
    output_names: list[str],  # Must have same length as function return
    input_otypes: Optional[dict],  # dict with keys matching function arg names
    output_otypes: Optional[dict],  # dict with keys matching output_names
    **function_kwargs,  # accepts values OR
   # other node channels (in which case it makes a connection, as now)
)
```

We'd still build the channels under the hood, but the users would never instantiate them. Subclassing node would then look something like

```python
class MyNode(Node):
    def __init__(self, **function_kwargs):
        super().__init__(
            self.my_function,
            ['first_out', 'second_out'],
            output_otypes = {
                'first_out': some_onto.SomeThing(),
                'second_out': some_onto.SomeOtherThing(),
            },
            **function_kwargs
       )

    @staticmethod
    def my_function(some_input: int, other_input: str) -> SomeThingClass, OtherThingClass:
        blah blah
        return a, b
```

This way it's not like sometimes you need `input_channels` and sometimes not, nor does `output_channels` change between `list[str]` and `list[ChannelTemplate]`.

samwaseda: ok I can live with this.

liamhuber: Super. When I get back to this, I'll do it that way and ping you again when it's in its new form.

samwaseda (commented):
One feature that would be really nice to have would be a nice representation of `__repr__`. Right now, what we see is `<pyiron_contrib.workflow.node.Node at 0x7f5366bf97b0>` for a node. Something like this:

```python
class Node:
    ...
    def to_dict(self):
        return {
            "label": self.label,
            "inputs": {
                tag: {
                    "type": str(channel.type_hint),
                    "connected": channel.ready,
                    "value": str(channel.value),
                }
                for tag, channel in self.inputs.channel_dict.items()
            },
            "outputs": {
                tag: {
                    "type": str(channel.type_hint),
                    "value": str(channel.value),
                }
                for tag, channel in self.outputs.channel_dict.items()
            }
        }
    
    def _repr_json_(self):
        return self.to_dict()
    
    def __repr__(self):
        return json.dumps(self.to_dict(), indent=2)
```

Obviously it's only a suggestion, so feel free to change it.

samwaseda:
```python
def f(a: int) -> int:
    return a + 1

def g(b: int) -> int:
    return b - 2

nd_f = Node(f, 'f')
nd_g = Node(g, 'g')
nd_f.inputs.a = nd_g.outputs.g
print(nd_f.inputs.a.ready)
```

Output:
```python
False
```

Is it possible that it's because `def connect` in `channels.py` only appends the connections but does not update `self.value`?

liamhuber:
> Is it possible that it's because `def connect` in `channels.py` only appends the connections but does not update `self.value`?

This is exactly why. I realized last night that that's what was bugging me about the pre-constructed nodes example, but didn't get a chance to fix it/add it to the todo list yet. I want to see if it introduces any weird side effects in the tests, but right now I plan to introduce updating-on-connection.

notebooks/workflow_example.ipynb
```diff
@@ -0,0 +1,493 @@
...
+    }
+   ],
+   "source": [
+    "def plus_minus_one(x):\n",
+    "    return x+1, x-1\n",
+    "\n",
+    "try:\n",
+    "    node = Node(plus_minus_one, (\"p1\", \"m1\"))\n",
+    "except TypeError:\n",
+    "    print(\"This won't quite work yet!\")"
```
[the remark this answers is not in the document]
liamhuber: That's fair. I don't think I'll fix it this PR though.

samwaseda:
I created a full fledged example of pseudo Murnaghan:

```python
from pyiron_contrib.workflow.node import Node
from pyiron_atomistics import Project
import numpy as np
from hashlib import sha1

from pyiron_atomistics.atomistics.structure.atoms import Atoms

pr = Project('WORKFLOW')

potential = '1995--Angelo-J-E--Ni-Al-H--LAMMPS--ipr1'

def get_energy(structure: Atoms, potential: str=potential, pr: Project=pr) -> tuple[float, float]:
    lmp = pr.create.job.Lammps(('lmp', sha1((structure.__repr__() + potential).encode()).hexdigest()))
    lmp.structure = structure
    lmp.potential = potential
    lmp.calc_static()
    lmp.run()
    return (lmp.structure.get_volume(per_atom=True) * 4)**(1/3), lmp.output.energy_pot[-1]

def get_strain(min_strain: float=-0.1, max_strain: float=0.1, n_points: int=11) -> np.ndarray:
    return np.linspace(min_strain, max_strain, n_points)

def get_min_lattice(strain: np.ndarray, pr: Project=pr) -> Atoms:
    v_lst, e_lst = [], []
    structure = pr.create.structure.bulk('Al', cubic=True)
    for s in strain:
        v, e = get_energy(structure.apply_strain(s, return_box=True))
        v_lst.append(v)
        e_lst.append(e)
    coeff = np.polyfit(v_lst, e_lst, 3)
    v_range = np.linspace(np.min(v_lst), np.max(v_lst), 1001)
    return structure.apply_strain(v_range[np.polyval(coeff, v_range).argmin()], return_box=True)

nd_min_lattice = Node(get_min_lattice, 'min_lattice')

nd_strain = Node(get_strain, 'strain')

nd_min_lattice.inputs.strain.connect(nd_strain.outputs.strain)
```

But I was not really sure how to make a workflow out of the inner loop, i.e. currently `get_energy` is not a node, because I didn't know how to make a variable number of nodes (in this case 11) into a workflow. Is there a better way to do it?

liamhuber:
> I created a full fledged example of pseudo Murnaghan:

Rad! Note that you can typically avoid passing `Project` as an argument; in the case of structures you just need to instantiate a structure factory, which is actually totally independent of projects anyhow (this is a good sign to me that Jan's work to pull the structure stuff entirely out of the module is on the right track). For jobs I've been playing with [totally deleting them and the project that made them](https://github.com/pyiron/pyiron_contrib/blob/5fe8dfedc3e13ddbefa548dec01a31a677ce2919/pyiron_contrib/workflow/nodes.py#L83) after the calculation is finished, so project can exist entirely under the hood there too. I've also been trying to work in a paradigm that one node creates a job and defines the engine-level properties (potential in this case, but kpoints etc in principle), then the node that actually calculates something [copies that job to do its work](https://github.com/pyiron/pyiron_contrib/blob/5fe8dfedc3e13ddbefa548dec01a31a677ce2919/pyiron_contrib/workflow/nodes.py#L61). We don't _have_ to do things this way, of course, but I've found both patterns useful. I think the next step is really to focus on lammps and just get them running totally HDF and database free with Jan's work.

> But I was not really sure how to make a workflow out of the inner loop, i.e. currently get_energy is not a node, because I didn't know how to make a variable number of nodes (in this case 11) into a workflow. Is there a better way to do it?

Yeah, this is a great question. I see three approaches:
1) Have a node that internally creates and runs (and IMO afterwards deletes) multiple jobs. There's no great reason a node can't be a bit heavy, and down the line we may even be able to assign multiple cores to node execution and let it parallelize under the hood (we'd need that later for, e.g. Vasp jobs anyhow, since we can't break that parallelization over multiple nodes). This is the easiest way, but maybe the least satisfying -- it works great when you're custom-building a node out of a function, but maybe results in too many very specific nodes if you try to make a new node class for each type of thing like this that you want.
2) Make a for-loop node ala [Unreal Engine's "blueprints"](https://docs.unrealengine.com/4.26/Images/ProgrammingAndScripting/Blueprints/UserGuide/FlowControl/ForLoop_Network.png). To keep the graph properly functional, I think we also need an "accumulator" node. I sketched these out while my kid was having a sick day, and I think we might need a `NotData` type that forces type checking to fail (thus influencing `ready`) regardless of whether the IO is typed or not. (He wanted to help so sorry it's not the most readable...)
3) Make some sort of "batching" paradigm, that automatically converts list-like collections of input data marked as "batched" into a loop during node execution. I do this [over in ironflow](https://github.com/pyiron/ironflow/blob/1b133b59de8437c958dbe7aa41f958ec36d0133c/ironflow/model/node.py#L348) and it's super sexy UX when you have a GUI because you just click a button to "batch" an input and all of a sudden your graph is running loops with no additional work on your end. The downside is that it massively complicates the already complex task of type checking, as you need to introduce a [whole new branch of logic](https://github.com/pyiron/ironflow/blob/1b133b59de8437c958dbe7aa41f958ec36d0133c/ironflow/model/dtypes.py#L110) to see what to accept once your channel is batched....

(1) will always be acceptable. At the moment I lean towards trying to get really tight flow control nodes to make (2) as easy as possible rather than porting over my work in ironflow for (3), just because of the type-checking headaches.

IMG_4533
Note: passing `nullify [NotData]` back into your own input makes the node `not ready` so that it won't update again until one of the other connected inputs updates. I think this for-loop is actually busted still, but the accumulator should work.

pyiron_contrib/workflow/io.py
```diff
@@ -0,0 +1,136 @@
...
+    is equivalent to
+    >>> some_io.some_existing_channel.connect(some_other_channel)
+    """
+    def __init__(self, *channels: Channel):
+        self.channel_list = [
+            channel for channel in channels if isinstance(channel, self._channel_class)
+        ]
+        self.channel_dict = DotDict(
+            {channel.label: channel for channel in self.channel_list}
+        )
```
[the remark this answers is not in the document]
samwaseda: Sorry I had done the reviews earlier, but apparently I forgot to submit them.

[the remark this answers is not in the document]
liamhuber: The original reason is that I got tired of writing `self.channel_dict.values()`, but you're absolutely right it's better with out. Changes made.

PR #603, opened 2023-02-28
Alternative to pyiron_base#1044

pmrv:
There's some higher order voodoo involved, but this should take care of the repetitive boiler plate you had to write for the nofile classes @jan-janssen .  I couldn't test in detail whether it works, because there seem to be problems in the rest of the code (I couldn't instantiate the original LammpsInteractiveWithoutOutput, e.g.; in this version you can instantiate it but it fails later in other code when you run it), so some debugging may be required, but I think the path to use it for the other classes should be clear.

### DIFF
--- a/pyiron_contrib/nofiles/wrapper.py
+++ b/pyiron_contrib/nofiles/wrapper.py
@@ -1,0 +1,65 @@
+from functools import wraps
+
+def wrap_without_files(cls, name, *methods):
+    """
+    Returns a subclass that defines the _interactive_disable_log_file flag and
+    overrides to_hdf, child_project and refresh_job_status to neither actually
+    write anything nor register the job with the database.
+
+    Additional methods (!, not method names) can be given to also wrap them so
+    that they are only called when the flag is not set.
+
+    Args:
+        cls (type): base class to derive from
+        name (str): name of the new class
+        *methods (functions): additional methods to be wrapped.
+
+    Returns:
+        type: subclass of cls that creates jobs that don't touch HDF or the
+        database
+    """
+
+    def init(self, project, job_name):
+        self._interactive_disable_log_file = False
+        super(cls, self).__init__(project=project, job_name=job_name)
+
+    @property
+    @wraps(cls.child_project)
+    def child_project(self):
+        if not self._interactive_disable_log_file:
+            return super(cls, self).child_project
+        else:
+            return self.project
+
+    @wraps(cls.to_hdf)
+    def to_hdf(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


pyiron_contrib/nofiles/wrapper.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def wrap_meth(meth):
+        """Return a method that calls its super() only if the flag is not set."""
+        @wraps(meth)
+        def wrapper(self, *args, **kwargs):
+            if self._interactive_disable_log_file:
+                return getattr(
+                        super(cls, self),
+                        meth.__name__
+                )(self, *args, **kwargs)
```
pmrv, line 61:
```suggestion
                )(*args, **kwargs)
```

jan-janssen:
> There's some higher order voodoo involved

That is the reason why I prefer https://github.com/pyiron/pyiron_base/pull/1044 . In the previous pull request, if somebody wants to add the `nofile` support to their class, all they have to do is add the `@to_hdf_decorator`, here it is much more complicated.

pmrv:
All they have to do with this is `LammpsNoFiles = wrap_no_files(Lammps)`, which I think is simple enough, since it's only the class developer that has to do this and even they don't have to touch the voodoo.  The main point is that with this function the change in functionality is localized to the code that needs, it cannot break anything in base and it doesn't incur more work for developers using it than pyiron_base#1044 .

jan-janssen:
> it cannot break anything in base and it doesn't incur more work for developers using it than pyiron_base#1044 .

The same applies to https://github.com/pyiron/pyiron_base/pull/1044 - only the classes which users want to be able without writing files need the decorator. So to me it is the same, just moved to a different location in the code. But as I already posted in https://github.com/pyiron/pyiron_base/pull/1044 based on the discussion on Monday, I moved on to develop the minimal version of pyiron in a separate module https://github.com/pyiron/pyiron_lammps . It now supports the sqs generation, structure optimization and elastic constant calculation I used in https://github.com/pyiron-dev/pyiron-no-file-access . The advantage is that it moves us closer to standalone quantum engines and it removes the dependence on h5py and sqlalchemy, so it is really the most light-weight version possible.

pmrv:
> So to me it is the same, just moved to a different location in the code. 

That's exactly the relevant difference.  By modifying this behavior only here, we can guarantee it cannot affect production code used by people not using this features.  

Anyway, I see you split your development out to the new repository, so feel free to use any solution over there you want, but this PR shows that continued prototyping of the nofile classes would have been very much possible without unreasonable effort.

PR #609, opened 2023-03-08
Tinybase: Reconstructing pyiron_base, just smaller

pmrv:
So after our long discussions for a while I've taken a stab at rewriting base with all the constraints in mind that we had talked about.  There are three example notebooks in the PR, but here I'll just go into the design a bit so we can discuss that. Mostly on my mind were:

1. Composability:
    1. Inputs/Outputs of workflows should be able to be composed by inheritance and 
    2. easily type checked 
    3. individual calculations ("nodes") should be easy to stack together
2. Flexibility:
    1. Generic workflows (murnaghan, minimizing, MD, etc.) should be possible without them knowing the details of their underlying "physics engines"
    2. Flexible resources allocation: Parallel nodes should be easy to distribute across threads/processes/etc. Serial nodes should respect their dependencies, ideally we should be able to combine such primitive nodes to build more complicated ones without loosing the ability to distribute tasks
    3. Not all computations need databases and file storage; they should be opt-in or -out
4. Resemblance to "old" pyiron:  we don't want to lose our growing user base and not everyone needs the added complexity of a super modular system

In `tinybase` most of that is achieved by breaking everything down and starting new from three fundamental concepts:
1. object oriented input/output,
2. calculation "nodes" using those and
3. "executors" that, well execute them.

Each node has a defined input and output class that allows stacking them together easily.  Nodes do not concern themselves with compute resources, they are just plain python functions.  Because input/output/nodes are can be pickled, the executors can just move them across processes and machines and move output back.  Node therefore don't need to know anything at all how they are executed.  Executors work with the `concurrent.futures` API.  I've checked that `dask` and `parsl` both also use this, so I'm optimistic that we could benefit a lot from the concurrency providers.

You could directly use just nodes if you are in a high throughput setting, but as it turns out you can also assemble them back into something that looks like `GenericJob`, but much simpler and more generic.  I've also sketched a very rudimentary database based on what @tnecnivkcots had been working on.

Currently this works:
1. Node primitives that allow to compose the workflows.  Probably not quite DAGs but close.  There is 
    1. function node that just executes python functions, 
    2. a list node that executes independent nodes (in parallel with the right executor)
    3. a series node that connects heterogeneous nodes in a fixed manner (think minimize then E-V curve)
    4. a loop node that connects homogeneous nodes in an iterative manner (think the actual minimizer, generic MD, etc.)
2. Some ASE specific nodes to have something to test all of this
3. A murnaghan workflow to test generic workflows.  It turns out that this one really benefits from the design, e.g. the whole murnaghan implementation is ~70 lines, but it can be combined with any node that outputs an energy, any combination of nodes that outputs an energy, it can run purely single threaded or scatter itself or its children to background processes or threads, it can be stored in the file system and the database or not, and any combination.
4. The reconstruction of pyiron_base:
    1. There's a generic storage layer plus adapter classes to use the current HDF files
    2. a simple database interface and a simple sqlalchemy implementation.
    3. `GenericTinyJob` which runs any node and saves it to database and filesystem

Caveats:
1. Except for the notebooks not a lot of documentation.
2. Very little error handling, if you do something wrong it won't tell you what. :|  I had the feeling a lot of this is in the graph based PR already, so I think we can cherry pick things between our prototypes.
3. Type checking between nodes and their input/output would be possible, but there's nothing implemented yet.
4. Not all the interfaces and classes are stable yet, especially the executors will need more work if we want to do remote execution and submitting things possible.
5. No tests. :[
6. Very little polish in terms of UI.  A lot of scattered imports and not always very convenient way of constructing things.  

@liamhuber I haven't yet looked in detail in to how you implemented your nodes, but I think despite the names they are a bit orthogonal.  I would think though, that it should be very easy to pull a graph of nodes from your frame work and make it behave like a node (or job) in `tinybase` or use the nodes here and turn them automatically into your nodes with channels and such.

I will update the docs


pmrv:
I forgot that the PR won't render notebooks, so here's some examples:

A murnaghan that sets up 500 strains, minimizes the atomic positions at each in a process pool, then plots the E-V curve.
```python
m = MurnaghanNode()
m.input.node = AseMinimizeNode()
m.input.node.input.calculator = MorsePotential()
m.input.node.input.max_steps = 100
m.input.node.input.output_steps = 10
m.input.node.input.ionic_force_tolerance = 1e-6
m.input.node.input.lbfgs()
m.input.structure = bulk("Fe", a=1.2)
m.input.set_strain_range(.5, 500)
m.input.child_executor = ProcessExecutor
exe = m.run(how='background')
exe.wait()
exe.output[0].plot()
```

This calculates fibonacci numbers in a background thread
```python
def calc_fib(n):
    import time
    n1 = n2 = 1
    for i in range(n):
        time.sleep(.1)
        x = n1 + n2
        n1 = n2
        n2 = x
    return x

f = FunctionNode(calc_fib)
f.input.kwargs['n'] = 100
exe = f.run(how='background')
exe._run_time, exe._collect_time
exe.output[0].result
```

Run an MD as an actual job, then show the trajectory
```python
j = GenericTinyJob(pr, 'md')
j.node_class = AseMDNode

j.input.structure = bulk('Fe', a=1.2, cubic=True).repeat(2)
j.input.calculator = MorsePotential()

j.input.steps = 100
j.input.timestep = 3
j.input.temperature = 600
j.input.output_steps = 20

j.run(how='foreground')
j.output.animate_structures()
```
```

File: pyiron_contrib/tinybase/container.py [hunk 1]
```diff
@@ -0,0 +1,179 @@
...
+        self.default_constructor = default_constructor
+        return self
+
+    def doc(self, text):
+        self.__doc__ = text
+
+class AbstractContainer(HasStorage, HasHDFAdapaterMixin, abc.ABC):
+    # TODO: this should go into HasStorage, exists here only to give one location to define from_ methods
+    @classmethod
+    def from_attributes(cls, name, *attrs, module=None, bases=(), **default_attrs):
```
liamhuber, line 100: Dynamic class creation makes me a little uncomfortable. Is there no more vanilla way to accomplish what you're looking for? My guess is that this is here for the syntactic sugar of making new input classes very quickly, but if we can find something, say, <33% longer but avoiding this magic I would take that hands-down.

pyiron_contrib/tinybase/executor.py
```diff
@@ -0,0 +1,201 @@
...
+        self._run_machine.step("finished",
+                status=self.status,
+                output=self.output,
+        )
+
+    def _run_finished(self):
+        pass
+
+    def run(self):
+        self._run_machine.step()
```
liamhuber, line 114: I find the run machine architecture is still not 100% transparent, but that this is already a big step up from the current schema -- nice job.

pyiron_contrib/tinybase/murn.py
```diff
@@ -0,0 +1,94 @@
...
+        "strains",
+        "node"
+)
+
+class MurnaghanInput(MurnaghanInputBase, ListInput):
+    def check_ready(self):
+        structure_ready = self.structure is not None
+        strain_ready = len(self.strains) > 0
+        node = self.node
+        node.input.structure = self.structure
```
liamhuber, line 32: This seems like an action, not a check...

pyiron_contrib/tinybase/murn.py
```diff
@@ -0,0 +1,94 @@
...
+
+class MurnaghanNode(ListNode):
+
+    def _get_input(self):
+        return MurnaghanInput()
+
+    def _get_output(self):
+        return MurnaghanOutput()
+
+    def _execute(self, output):
```
liamhuber, line 80: Why does `_execute` take _output_ as an argument?? Something has gone sideways with the naming scheme here.

pyiron_contrib/tinybase/murn.py
```diff
@@ -0,0 +1,94 @@
...
+        output.base_structure = self.input.structure
+        ret = super()._execute(output)
```
liamhuber, line 82: `ret` is not used? ...are we passing around a (giant?) mutable data object in `output` and implicitly modifying it through different layers of `_execute`?

pyiron_contrib/tinybase/storage.py
```diff
@@ -0,0 +1,264 @@
...
+
+    @abc.abstractmethod
+    def _store(self, storage):
+        pass
+
+    def _store_type(self, storage):
+        storage["NAME"] = self.__class__.__name__
+        storage["MODULE"] = self.__class__.__module__
+        # DESIGN: what happens with objects defined in different parts of pyiron{_contrib,_atomistics,_base}?  Which
+        # version do we save?
```
liamhuber, line 209: If people are using the conda installs, storing only the module that actually owns the class should be fine, since we always hard-limit our version numbers on the dependencies.

If it's not a version but some random commit, and they're working with bleeding edge stuff that is somehow interdependent, then I'm not so sure what will happen...

liamhuber (commented):
Cool! I haven't looked at the notebooks yet and will go through it in more detail again later, but I wanted to give you quick turnaround on some first feedback.

> I haven't yet looked in detail in to how you implemented your nodes, but I think despite the names they are a bit orthogonal.

Yes. There might be some painful merging in handling IO, but overall I think there is a fair amount of orthogonality that should make the two prototypes mutually-beneficial.

>>  Type checking between nodes and their input/output would be possible, but there's nothing implemented yet

>> No tests. :[

In particular I've got both, which should be complementary/stabilizing.

I don't handle storage at all, and treat execution quite naively, so I see a strong benefit from your pickling and executors.

> Node primitives that allow to compose the workflows.

So if I see correctly, you basically have nodes that take as input other nodes. I would push back relatively strongly on this, at least if we want to take any strong stance that both categories are "just nodes". It's tempting from a code perspective, but I run into huge headaches when I try to think about how to represent it, either in a GUI or on a whiteboard.

If you pass in node _instances_ (1) how do you draw (in a GUI on a diagram) that a node is passing its _entire self_ into another node's input? You basically get stuck with _every_ node needing a `self` output. Then (2) once it's in there, are you always doing big deepcopies of the entire node, or are you using mutable data? Deep copying will be pretty safe (if your entire node is copyable, which it sounds like yours are), but I worry about the performance hit and overkill of passing around more data than is needed; and passing mutable data around graphs is just super risky.

If you pass a node _class_ then you need to re-expose all the child input on the now-owning super-node. This is doable on the code side, although it involves a bunch of dynamic work. You can do the same dynamic work in a GUI, but it's _way_ more work.

We definitely need some way of doing the flow control that you have, and I am open to doing it with an object that takes nodes as input, but then I would be extremely careful about asserting that these new flow control objects are themselves also nodes.

Lastly, is there some high-level sketch to help understand your new work? Since you reproduce all of base, including database and storage stuff, it's a lot more complex that what I have over on the `graph` branch. Nonetheless, there it can be really easily understood as `Channel<IO<Node<Workflow`, where the leftmost class can always be understood and tested in terms of only itself and the classes to the left of it, and each class owns instances of the class to its immediate left. Is there a similar framework I can use to parse this code?

pmrv:
> So if I see correctly, you basically have nodes that take as input other nodes. I would push back relatively strongly on this, at least if we want to take any strong stance that both categories are "just nodes". It's tempting from a code perspective, but I run into huge headaches when I try to think about how to represent it, either in a GUI or on a whiteboard.

I think they must be nodes to allow easy composition of things.  Say I want to make a minimizer and I have an engine that gives me forces and a loop primitive (that is not a node itself).  I can write the minimize algorithm generically and had it together with the engine to the loop primitive.  But how do I run this object now generically?  Maybe the framework can turn the loop primitive into a node for me that I can run, but at that point it could have been a node to start with. 

If the fear is purely related to GUIs then a solution could be to have two separate connections: data flow as we have it currently and control flow.  Make the control flow connection connect to a whole node instead a "faked" self output.  If the lines showing these to connections are visually distinct I don't see a problem with having them both exposed. 

Anyway I don't think there must be a one-to-one mapping between nodes in the GUI graph and actual compute nodes.  It would be easy for developers if you can turn compute nodes into the GUI nodes quickly, but graph users don't have to care how their graph runs.



> Lastly, is there some high-level sketch to help understand your new work? Since you reproduce all of base, including database and storage stuff, it's a lot more complex that what I have over on the `graph` branch. Nonetheless, there it can be really easily understood as `Channel<IO<Node<Workflow`, where the leftmost class can always be understood and tested in terms of only itself and the classes to the left of it, and each class owns instances of the class to its immediate left. Is there a similar framework I can use to parse this code?

Nowhere explicit yet, but I guess you can draw it up in a similar way as `AbstractContainer<AbstractInput/AbstractOutput<AbstractNode<Executor<TinyJob`.  Nodes are the ones who do all the composition.  An executor just turns a list of nodes into a list of outputs, that's really it.

A tiny job just bundles these together and logs it on the filesystem and the database.  This additional functionality comes mostly from the `ProjectInterface`, which is sort of an abstract project.  It provides two important things: 1) black box access to `GenericStorage` (or `ProjectHDFio` pyiron_base speak) and 2) a generic database interface.

1) is nice because it leaves the project in complete control over the location and type of storage.  In the current setup it'd really easy to have all jobs stored in one file or completely in memory.
2) is just the continuation of `IsDatabase` from pyiron_base.  I tried to use it first, but it's just extremely annoying to use outside of base.

pyiron_contrib/tinybase/murn.py
pmrv: This is just done so that `node.check_ready` can be called to so if the "sub-node" is ready. The whole input checking logic is quite weak though, so maybe there's a better way for this.

pyiron_contrib/tinybase/container.py
pmrv: Sure,

```python
class MyInput(AbstractContainer):
    field1 = StorageAttribute()
    field2 = StorageAttribute().type(int).default(42)
```
is the same as the corresponding `from_attributes` call.  I just got lazy prototyping.  I was thinking it'd be cute to be able to make an input class from a `dict` of default and this is where `from_attributes` comes from, but I didn't follow through with it.

pyiron_contrib/tinybase/executor.py
pmrv: Yeah, the relationship between the executors and the run state machine and how the state machine handles and exposes `_data` is also not quite solidified for me.  You'll see a few places still where `TinyJob` needs to reach inside all of this.

pyiron_contrib/tinybase/murn.py
liamhuber: > which is what everyone has to implement

So I think one of the things I'm running up against here is the volume of moving pieces when defining a new node. If I understand correctly, first new input and output classes need to be defined (inheriting from their respective `Abstract`s) -- what I like about this is the ability to leverage composition. 

Then the user defines `_execute`, which is the real meat of the node, this takes an instance of the output as an argument; internally the user is then expected to leverage the attributes of `self.input` to modify the attributes of `output`. The `execute` wrapper for `_execute` then makes sure the modified output instance gets returned. 

The main thing I don't like about this is how implicit everything is when defining the key aspect of node behaviour -- when defining `_execute` the dev needs to refer back to their IO classes all the time; they might be able to get decent variable name hints for `output` if they make sure to type hint `_execute(self, output: MyOutputClass)`, but they will never be able to get good hints for `self.input` from their IDE because it is (perfectly correctly) type hinted at the parent class level as `AbstractInput`. 

Additionally it looks like the output class...can have default values? `AbstractOutput.from_attributes(foo=list)` uses the `list` constructor to make an empty one? Then I see big danger for a dev making their output class, forgetting to populate parts of it, and silently getting totally weird returned values.

> but if _execute produces the output itself instead of getting it passed, then `execute` would have to do a type check afterwards

I don't think I agree with that; why would we type check outputs? If I write `def foo(x: int) -> str: return x`, that's _my_ fault! 😜 In the same way, if they say they're returning an instance `MyMurnOutput` and return `MyFooBarOutput`, I don't see the need for us to catch them.

pyiron_contrib/tinybase/murn.py
liamhuber: I'm afraid I'm still not sure I follow, but I'll keep reading the code.

The fact you invoked "sub-node readiness" does, however, add to my complaint about code-smelliness with having nodes take nodes as input....

pyiron_contrib/tinybase/container.py
liamhuber: I think the crux of the issue here is that I'm somewhat torn between the ability to do stuff my composition when we explicitly define IO classes, and the elegance/stability of generating IO automatically from the actual behaviour of the node.

I think there is a direct analogy here where I _used_ to (and, e.g. ryven3 does) have templates for node IO for each of the individual IO variables. You could package these templates together and build them by composition, in exact analogy to what you do here. Just like here, you wind up needing to explicitly state type variables, defaults, etc. Then your trick with `from_attributes` becomes helpful syntactic sugar for making this all easier to type.

What @samwaseda really pushed me to over in the graph PR is instead to _automatically_ scrape as much of this (arg name, type(optional), default(optional)) from the functional definition of the node behaviour itself. (Note: the definition is functional, but if you pass around mutable data as arguments you still have the freedom/danger of non-functional graphs.)

Maybe there is a way to think about getting some sort of function composition going to have the best of both worlds, because after he pushed me to go that route, I would up reeeaaally liking Sam's information-extraction approach.

pyiron_contrib/tinybase/container.py
```diff
@@ -0,0 +1,179 @@
...
+            return obj.storage[self.name]
+        except KeyError:
+            if self.default_constructor is not None:
+                ret = obj.storage[self.name] = self.default_constructor()
+                return ret
+            else:
+                return None
+
+    def __set__(self, obj, value):
+        if isinstance(value, self.value_type):
```
liamhuber, line 63: So, painfully, this won't work, at least not robustly.

The problem is that you can't use `isinstance` with `typing` generics. With a modern enough version of python you can get `isinstance("a", str | int)` and `isinstance("a", typing.Union[str, int])` both working, but there are other extremely useful and necessary type hints like `Literal` and `Callable` which throw an error, e.g. `isinstance("a", Literal["a"])`.

I wound up leaning on the `typeguard` package to handle both cases. It's doable, it's just more verbose than this check.

pyiron_contrib/tinybase/container.py
pmrv: I think there's a good middle.  Actually `from_attributes` would work nicely together with getting names and types from the function definition, so you could generate these on the fly quite easily.  What you lose  I think is then the ability to compose it nicely and safely.  E.g. a Murnaghan needs to know that it can just stick a structure into a node and get an energy out.   In a notebook I agree its very convenient to derive it from a annotations, but for more long lived code I'd prefer a bit more OOP.  I do think we can have both by making the compatibility check between take into account first the class relations and then whether matching attributes are defined like you do in the graphs.

pyiron_contrib/tinybase/container.py
pmrv: Ah, value type is in this case really a concrete class not a type hint.  I don't see a fundamental problem with using `typeguard` or such to allow it, though.  More likely we'll also need range checks though, so I think at some point we'll just have to let devs provide functions to check values.  I think this is pretty similar to how we had talked about `traitlets`.  I'm not so worried that this might become a bit tricky, because we'll only need to define such a "general type check once and export it so that node devs can use it.

pyiron_contrib/tinybase/murn.py
pmrv: > The main thing I don't like about this is how implicit everything is when defining the key aspect of node behaviour -- when defining `_execute` the dev needs to refer back to their IO classes all the time; they might be able to get decent variable name hints for `output` if they make sure to type hint `_execute(self, output: MyOutputClass)`, but they will never be able to get good hints for `self.input` from their IDE because it is (perfectly correctly) type hinted at the parent class level as `AbstractInput`.

I'm not sure I see the problem.  Is the need the refer back to their input/output not part of any solution that we'd come up with?  If you define the node input implicitly based on the function arguments they'd have everything visible in their definition of `_execute`, but it'd be very easy to forget some parts of the arguments of all the input they ought to support to be general, so they still need to check if the function that they wrote actually match the generic interface by... looking at the corresponding generic input class.

As for type hinting of input, I think we can come up with something to make that work, e.g. node devs annotating their `_get_input` with the input type.



> Additionally it looks like the output class...can have default values? `AbstractOutput.from_attributes(foo=list)` uses the `list` constructor to make an empty one? Then I see big danger for a dev making their output class, forgetting to populate parts of it, and silently getting totally weird returned values.

Fair, but imo solvable by working on the `from_attributes` or `StorageAttribute` builders.



> > but if _execute produces the output itself instead of getting it passed, then `execute` would have to do a type check afterwards
> 
> I don't think I agree with that; why would we type check outputs? If I write `def foo(x: int) -> str: return x`, that's _my_ fault! 😜 In the same way, if they say they're returning an instance `MyMurnOutput` and return `MyFooBarOutput`, I don't see the need for us to catch them.

Ja, here I very strongly disagree.  For things that are defined in the notebook interactively you are right that it doesn't matter much, but for "official" nodes that live really in one of the pyiron modules I'd like to give really strong guarantees that they work as advertised.  Otherwise the same problems that currently haunt the generic methods like `calc_*` in atomistics will come back, i.e. you explicitly need to check that sub classes take the same arguments with the same semantics.

pyiron_contrib/tinybase/murn.py
samwaseda: I'm ok with @pmrv 's solution. I agree that the current name is quite clumsy -> it should be something that implies that it prepares the simulation based on the input and returns whether the setup is is correct or not (no idea what it should be called though XD).

At the end of the day, what must be clear is the following routine:

User input -> Compilation -> Execution -> Output collection -> User output

And `check_ready` would correspond to the compilation layer in this flow. In the current pyiron implementation, we have several functions for this, i.e. `check_setup`, `validate_ready_to_run` and `write_input`. But as I already said, I'm also ok with having only one function for this.

notebooks/tinybase/Basic.ipynb
```diff
@@ -0,0 +1,1183 @@
...
+   "source": [
+    "exe._run_time, exe._collect_time"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a6cbcd0e-2a04-4e93-add5-0f423fc1fa69",
+   "metadata": {},
+   "source": [
+    "## We don't actually have to use an executor if we just want a result"
```
samwaseda, line 294: This example is difficult to understand, because it doesn't tell how to make a comparable output as in the subsection above.

pyiron_contrib/tinybase/node.py
```diff
@@ -0,0 +1,404 @@
...
+        pass
+
+    @property
+    def input(self) -> AbstractInput:
+        if self._input is None:
+            self._input = self._get_input()
+        return self._input
+
+    @abc.abstractmethod
+    def _get_output(self) -> AbstractOutput:
```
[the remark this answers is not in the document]
pmrv: Yes, that's correct.  Agree that the name can be better. Maybe `_initialize_output`?

pyiron_contrib/tinybase/node.py
```diff
@@ -0,0 +1,404 @@
...
+
+    @classmethod
+    def _restore(cls, storage, version):
+        node = cls()
+        node._input = pickle_load(storage["input"])
+        return node
+
+FunctionInput = AbstractInput.from_attributes("FunctionInput", args=list, kwargs=dict)
+FunctionOutput = AbstractOutput.from_attributes("FunctionOutput", "result")
+class FunctionNode(AbstractNode):
```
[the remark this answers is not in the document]
pmrv: Yeah fair, my thinking now is that I make an adapter somehow to go from Liam's nodes to mine and then just use his code.

samwaseda:
...because I was following the steps suggested by ChatGP to copy the commits of the potential branch from pyiron_atomistics, which didn't quite do what it was supposed to do...

pyiron_contrib/tinybase/lammps.py
```diff
@@ -0,0 +1,189 @@
...
+        potential.write_file(
+            file_name="potential.inp", cwd=self.input.working_directory
+        )
+        potential.copy_pot_files(self.input.working_directory)
+
+        control = LammpsControl()
+        if self.input.calc_type == "static":
+            control.calc_static()
+        else:
+            assert "Cannot happen"
```
samwaseda, line 89: Isn't this line meaningless? Is `raise AssertionError("Cannot happen")` meant?

pyiron_contrib/tinybase/container.py
```diff
@@ -0,0 +1,186 @@
...
+    this property before setting it, `None` is returned.
+
+    It's possible to modify the default value and the accepted type of values by using the builder-style :meth:`.type`
+    and :meth:`.default` methods.
+
+    >>> class MyType(HasStorage):
+    ...     a = StorageAttribute()
+    ...     b = StorageAttribute().type(int)
+    ...     c = StorageAttribute().default(list)
+    ...     d = StorageAttribute().default(lambda: 42).type(int)
```
samwaseda, line 30: I find this this initialisation difficult to follow. I would prefer `StorageAttribute().default(42).type(int)` and inside `def default(self)` I'd check if it's a type or a value.

samwaseda: Actually I might totally scrap the callable option and force the user to set a default value, i.e. if it's a list it must be `StorageAttribute().default([])`

pyiron_contrib/tinybase/lammps.py
pmrv, line 89:
```suggestion
            assert False, "Cannot happen"
```

samwaseda: From what I know, [assert is not compatible with PEP8](https://stackoverflow.com/questions/16065482/how-to-format-a-python-assert-statement-that-complies-with-pep8)

samwaseda: I guess what makes sense is:

```python
assert self.input.calc_type == "static"
control.calc_static()
```

Otherwise `assert False` looks super clumsy...

pmrv: There is a typo :S , but otherwise the assert statement and raising the error is equivalent.

pyiron_contrib/tinybase/container.py
pmrv: Ja, I agree it's much nicer to have to provide just the value instead of the function.  Naively it won't work to write `StorageAttribute().default([])` because every instance of this class would share the *same list instance*, i.e. if one class instance modifies its property it would be reflected in another (which we don't want).  I'll see if I can sneak in a copy somewhere though.

pyiron_contrib/tinybase/lammps.py
pmrv: Fair enough, I'll change that.

pmrv: > From what I know, [assert is not compatible with PEP8](https://stackoverflow.com/questions/16065482/how-to-format-a-python-assert-statement-that-complies-with-pep8)

Not sure I follow, the link just says it's difficult to fit assert statements within 80 chars (which we don't adhere to anyway), but also gives solutions.

pyiron_contrib/tinybase/container.py
samwaseda: Can't we just do

```python
self.default_constructor = lambda: default_constructor
```

And

```python
ret = obj.storage[self.name] = self.default_constructor()()
```

Not sure if this gives a different result though.

pyiron_contrib/tinybase/lammps.py
samwaseda: yeah to be honest I also never understood. @sudarsan-surendralal pointed it out some years ago because apparently some tests failed because of `assert`, but I personally also like it and use it in multiple places.

pyiron_contrib/tinybase/container.py
pmrv: Sorry, I dropped the ball here for a bit.  I'm not sure what the double call is supposed to do.  I think I'll just `copy.deepcopy()` whatever the user gives in.

pmrv: I've added both options now, and `default` does the intuitive thing now.

PR #618, opened 2023-03-21
change np.int to np.int64 to fix pacemaker job with numpy >=1.24

srmnitc:


srmnitc:
Merging this without review since it is a really small change and urgently needed for the upcoming workshop. All tests pass.

PR #633, opened 2023-04-12
Introduce control channels

liamhuber:
Workflow nodes now have a space for signal channels that control callbacks. By default they have a `run` channel (which has the node's `run` method as a callback) and a `ran` channel (which triggers at the end of the `run` method). Adding more is possible, so we should be able to get more complex conditional flow as discussed (among many other things) [here](https://github.com/pyiron/ironflow/issues/188#issuecomment-1504129976) and as seen in [Unity's blueprints scheme](https://docs.unrealengine.com/4.26/en-US/ProgrammingAndScripting/Blueprints/UserGuide/FlowControl/). (Edited to add hyperlinks)

Here is the new content of the demo:

# Flow control

By default, when a node runs and updates its output, this triggers outputs in all downstream connections. This is useful when all your node functions are small and light, but there may come times when you want something other than this simple "push" flow.

In addition to input and output data channels, nodes also have "signal" channels available. Input signals are bound to a callback function (typically one of its node's methods), and output signals trigger the callbacks for all the input signal channels they're connected to.

Standard nodes have a `run` input signal (which is, unsurprisingly, bound to the `run` method), and a `ran` output signal (which, again, hopefully with no great surprise, is triggered at the end of the `run` method.)

Below is a super simple example of how these signal channels can be used to delay execution and manually control flow (you will need `from pyiron_contrib.workflow.node import node`):

```python
@node("y")
def linear(x):
    return x

@node("z")
def times_two(y):
    return 2 * y

l = linear(x=1)
t2 = times_two(
    y=l.outputs.y, update_on_instantiation=False, run_automatically=False
)
print(t2.inputs.y, t2.outputs.z)
>>> 1, None
```

Now the input of `t2` got updated when the connection is made, but we told this node not to do any automatic updates, so the output has its uninitialized value of `None`.

Often, you will probably want to have nodes with data connections to have signal connections, but this is not strictly required. Here, we'll introduce a third node to control the execution of `t2`.

Note that we have all the same syntacic sugar from data channels when creating connections between signal channels.

```python
@node("void")
def control():
    return

c = control()
t2.signals.input.run = c.signals.output.ran
c.run()
print(t2.outputs.z.value)
>>> 2
```

### DIFF
--- a/notebooks/workflow_example.ipynb
+++ b/notebooks/workflow_example.ipynb
@@ -29,7 +29,7 @@
     {
      "data": {
       "application/vnd.jupyter.widget-view+json": {
-       "model_id": "9aa721332f2644d5a30634bf7dfac067",
+       "model_id": "ceee180feeb547a0b959ccacf8aef2ae",
        "version_major": 2,
        "version_minor": 0
       },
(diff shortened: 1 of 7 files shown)


liamhuber:
Merging in main killed the pip check with something unrelated to what I'm doing here:

```
cfn-lint 0.75.1 has requirement aws-sam-translator>=1.60.1, but you have aws-sam-translator 1.55.0.
```

liamhuber:
> Merging in main killed the pip check with something unrelated to what I'm doing here:
> 
> ```
> cfn-lint 0.75.1 has requirement aws-sam-translator>=1.60.1, but you have aws-sam-translator 1.55.0.
> Error: Process completed with exit code 1.
> ```

@niklassiemer, the only place I see AWS mentioned is in the S3 business -- are either of the above libraries meaningful to you?

PR #637, opened 2023-04-24
Option to store/download working directory

niklassiemer:
The downloading is not into the right place to reconstruct the full job directory. However, right now also the names etc. are not pyrionic, nothing is stored in the database and such. 
Has to be cleared in conversion to a realistic `ArchiveProject`.

### DIFF
--- a/pyiron_contrib/generic/coscineIo.py
+++ b/pyiron_contrib/generic/coscineIo.py
@@ -305,7 +305,12 @@ class CoscineResource(StorageInterface):
         self._resource.upload(filename, file, _meta_data)
 
     def upload_job(
-        self, job: pyiron_base.GenericJob, form=None, update_form=False, dois=None
+        self,
+        job: pyiron_base.GenericJob,
+        form=None,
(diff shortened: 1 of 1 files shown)


pyiron_contrib/generic/coscineIo.py
```diff
@@ -326,24 +332,54 @@ def upload_job(
             form["DOIs"] = dois
 
         self.upload_file(job_file, form)
+        if upload_working_dir:
+            if not job.is_compressed():
+                job.compress()
+            self.upload_file(
+                os.path.join(job.working_directory, job.job_name + ".tar.bz2"), form
```
pmrv, line 339: `pyiron_base` has a function that constructs the archive name, best to use it.  (I forgot the name right now, but it's in the same module as the compress/decompress functions)

pmrv: Ah, sorry that was part of https://github.com/pyiron/pyiron_base/pull/969 which I didn't get around to merge.

PR #644, opened 2023-04-25
Update base

niklassiemer:


liamhuber:
@niklassiemer I'm getting a similar error on #645, but you seem to make it a bit farther. When I hit line 716 (`looking for...`) I get a seg fault. You make it a bit farther and we actually see the error `RuntimeError: Invalid spec, no package name found: NULL>`. 

I don't see any human-readable issues with .ci_support/environment.yml (nor ./setup.py). I want to actually do some real work so I'm just going to stack my PRs and ignore this for now, but I should have a chance to look at it on Thursday/Friday. Ping me if you find a solution in the meantime.

niklassiemer:
Thanks for reporting - I also did not find a flaw in the files and wondered what is going on. However, I won't have time to dig into this this week.

niklassiemer:
Already in from dependabot

PR #647, opened 2023-04-28
i-pi job update

raynol-dsouza:
PiMD jobs now only collect the beads' centroid positions and forces instead of the positions and forces of *all* the beads.


liamhuber:
@niklassiemer I didn't get to working on our conda problem, but @raynol-dsouza's PR got an even more informative error message than either of us!!! It's late on a Friday for me so I've got to stop working, but I'm super optimistic that we now have enough data to resolve this. It's very weird that our three PRs terminally crashed at such different stages of this error though... if we'd just gotten this error message to start with it would have been much more clear how to proceed....

```
├─ pyiron_atomistics 0.2.66**  is installable and it requires
│  └─ scipy >=1.10.1  with the potential options
│     ├─ scipy [1.10.1|1.8.0|1.8.1|1.9.0|1.9.1] would require
│     │  └─ python >=3.10,<3.11.0a0 , which conflicts with any installable versions previously reported;
│     ├─ scipy [1.10.1|1.8.0|1.8.1|1.9.0|1.9.1] would require
│     │  └─ python >=3.8,<3.9.0a0 , which conflicts with any installable versions previously reported;
│     ├─ scipy 1.10.1, which can be installed;
│     └─ scipy 1.10.1 would require
│        └─ python >=3.11,<3.12.0a0 , which conflicts with any installable versions previously reported;
└─ scikit-image 0.20.0**  is uninstallable because there are no viable options
   ├─ scikit-image 0.20.0 would require
   │  └─ python >=3.10,<3.11.0a0 , which conflicts with any installable versions previously reported;
   ├─ scikit-image 0.20.0 would require
   │  └─ python >=3.11,<3.12.0a0 , which conflicts with any installable versions previously reported;
   ├─ scikit-image 0.20.0 would require
   │  └─ python >=3.8,<3.9.0a0 , which conflicts with any installable versions previously reported;
   ├─ scikit-image 0.20.0 would require
   │  └─ scipy >=1.8,<1.9.2  but there are no viable options
   │     ├─ scipy [1.10.1|1.8.0|1.8.1|1.9.0|1.9.1], which cannot be installed (as previously explained);
   │     ├─ scipy [1.10.1|1.8.0|1.8.1|1.9.0|1.9.1], which cannot be installed (as previously explained);
   │     └─ scipy [1.8.0|1.8.1|1.9.0|1.9.1] conflicts with any installable versions previously reported;
   └─ scikit-image 0.20.0 conflicts with any installable versions previously reported.
```

liamhuber:
```
├─ scikit-image 0.20.0 would require
   │  └─ scipy >=1.8,<1.9.2  but there are no viable options
```

So on the [scipy requirements](https://github.com/scikit-image/scikit-image/blob/main/pyproject.toml), I see the lower pin to `>=1.8`, but I don't see the upper pin anywhere....

Ok, that's just because the git head is not 0.20.0 anymore. Indeed, on 0.20.0 [the requirement is there](https://github.com/scikit-image/scikit-image/blob/v0.20.0/requirements/default.txt).

I'm going to check if 0.21.0 (which _only_ has the lower bound) is available on conda yet... Nope. Not showing on [conda releases](https://anaconda.org/conda-forge/scikit-image) and there's no PR for it yet [on their feedstock](https://github.com/conda-forge/scikit-image-feedstock/pulls).

However, as soon as it's released, bumping our dependency to 0.21.0 should resolve all the currently failing tests.

raynol-dsouza:
@liamhuber thank you for getting this sorted! Shall I go ahead and merge this PR?

pyiron_contrib/atomistics/ipi/ipi_core.py
```diff
@@ -136,23 +137,47 @@ def collect_props(self):
...
+            snap_lines = lines[start:stop - 1]
+            for i, l in enumerate(snap_lines):
+                split_l = l.split()
+                if i != 0:
+                    temp_list.append([float(j) for j in split_l[1:]])
+                else:
+                    abc.append([float(j) for j in split_l[2:5]])
+                    ABC.append([float(j) for j in split_l[5:8]])
+            start = stop
+            traj.append(temp_list)
```
liamhuber, line 164: This is a little hard to understand. Could you refactor it out into a well-named method? (It looks like the two previous lines defining `abc` and `ABC` would need to be pulled out at the same time.

I also think that the loop body could be simplified a bit, although do please double check that what I have here is actually doing the same thing, since I don't have data to test on:
```suggestion
        traj = []
        for i, stop in starts[1:]:
            start = starts[i - 1]
            snap_lines = lines[start:stop - 1]
            
            zeroth_line_values = [float(n) for n in snap_lines[0].split()]
            abc.append(zeroth_line_values[2:5])
            ABC.append(zeroth_line_values[5:8])
            # TODO: Comment on what is being grabbed by these slices?      
            
            temp_list = []
            for j, l in enumerate(snap_lines[1:]):
                temp_list.append([float(n) for n in l.split()[1:]])
            traj.append(temp_list)
```

What I like about this is that it gets rid of the `if` clause, making the interior loop super simple s.t. the nested loop is overall quite straightforward.

pyiron_contrib/atomistics/ipi/ipi_core.py
```diff
@@ -136,23 +137,47 @@ def collect_props(self):
...
         f.close()
-        self.custom_output.cell_abc = np.array(abc)
-        self.custom_output.cell_ABC = np.array(ABC)
+        return np.array(abc), np.array(ABC), np.array(traj)
+
+    def collect_trajectory(self):
+        #digits = "{0:0" + str(len(str(self.custom_input.n_beads))) + "}"
+        #f = open(self.working_directory + '/ipi_out.pos_' + digits.format(0) + '.xyz')
```
liamhuber, line 170: If these are really not needed anymore just kill them. They'll always be there in the git history if it becomes absolutely necessary to go back and look at them.

liamhuber (commented):
@raynol-dsouza, lgtm. I have no idea how these jobs actually run or what the rest of their code looks like, but I trust you that it's performing as expected. 

I made a few minor comments for (hopefully) improving readability.

pyiron_contrib/atomistics/ipi/ipi_core.py
raynol-dsouza: Needed a bit of modification to run properly. Added documentation.

pyiron_contrib/atomistics/ipi/ipi_core.py
```diff
@@ -139,33 +139,36 @@ def collect_props(self):
...
+            # second line contains the cell data. Collect it.
+            zeroth_line_values = [n for n in snap_lines[0].split()]
+            cell_abc.append(np.array(zeroth_line_values[2:5], dtype=float))  # collect cell box edges
+            cell_ABC.append(np.array(zeroth_line_values[5:8], dtype=float))  # collect cell box angles
+            # the remaining lines of the snapshot contain the positions/forces. Collect them.
+            temp_list = []
+            for l in snap_lines[1:]:
+                temp_list.append([float(n) for n in l.split()[1:]])  # first column is the species. Ignore it.
+            trajectory.append(temp_list)
             start = stop
```
liamhuber, line 167: I still find this formulation really difficult to read. Could you switch it with `start = starts[i-1]` and move it up to the top?

liamhuber (approved):
I still have one nit about `start = stop`, but otherwise looks better!

pyiron_contrib/atomistics/ipi/ipi_core.py
raynol-dsouza: `starts[i-1]` will become `starts[-1]` when `i=0`. I kept it this way to avoid this. Looking at it again, I think `start=starts[i]` at the beginning of the loop should work.

liamhuber: > Looking at it again, I think start=starts[i] at the beginning of the loop should work.

Good catch! Yeah, that should be perfect

PR #648, opened 2023-04-29
Make normal nodes slow

liamhuber:
I.e. they should default `runs_on_updates` and `updates_on_instantiation` both to `False`.

All the documentation and examples are updated accordingly, and the library of nodes makes as many as possible `SingleValueNodes`.


liamhuber:
This stacked PR updates the docstrings and example notebook, so I don't want to merge it's parent without it. Merging into parent now.

PR #658, opened 2023-05-02
Qol workflow reprs

liamhuber:
- Channel values are now stringified before being placed in the `to_dict` dictionary, which avoids errors when trying to view, e.g., channels that hold an `Atoms` object. A different solution will be needed for actual serialization, but this works fine for display purposes.
- `Workflow` and `IO` somehow aren't getting their pretty JSON representation triggered, even though `Node` and `Channel` do and they all just inherit this functionality from `HasToDict`. Until I can figure out why and fix it, I added a method `json_repr()` to force this behaviour.

### DIFF
--- a/pyiron_contrib/workflow/has_to_dict.py
+++ b/pyiron_contrib/workflow/has_to_dict.py
@@ -10,6 +10,5 @@ class HasToDict(ABC):
     def _repr_json_(self):
         return self.to_dict()
 
-    @property
     def info(self):
         print(dumps(self.to_dict(), indent=2))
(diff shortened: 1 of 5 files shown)


pyiron_contrib/workflow/has_to_dict.py
```diff
@@ -10,6 +12,8 @@ def to_dict(self):
     def _repr_json_(self):
         return self.to_dict()
 
-    @property
     def info(self):
         print(dumps(self.to_dict(), indent=2))
+
+    def repr_json(self):
```
niklassiemer, line 18: There shoud be a __repr_json__ or similar method used in notebooks to Display the object automatically as json

PR #675, opened 2023-05-10
Add a doc example for workflow inputs and outputs

### LINKED ISSUE #674: How can I get the output in workflow?
I followed the example in the doc string:

```python
def fnc(x=0): return x + 1
n1 = Node(fnc, "x", label="n1")
wf = Workflow("my_workflow", n1)
```

While it explains myriads of ways to set inputs, it nowhere explains how to get the output, and I'm struggling to get it.

liamhuber:
Closes #674


liamhuber:
Don't worry about test failures; github is currently experiencing issues. [Updates here](https://www.githubstatus.com)

liamhuber:
@samwaseda bump.

samwaseda (commented):
I have the feeling that the fundamental examples (such as the ones showing how to access output) should be contained within a class and need not import other classes, because a) it makes it easier for the user to read; b) it makes it easier for the developer to keep doc strings up to date. I find it good that there’s this MD example in this doc, but I might prefer to have this new part executable without importing other classes (I presume it’s possible, right?)

liamhuber:
> I have the feeling that the fundamental examples (such as the ones showing how to access output) should be contained within a class and need not import other classes, because a) it makes it easier for the user to read; b) it makes it easier for the developer to keep doc strings up to date.

Hmm, I like the sentiment. I do think it's fair to import `Node` though, since that's the object that gets added to workflows. It _is_ available directly on the workflow now (`wf.add.Node`), but I still think it's nice to see a wild node get added. Does that feel like a fair compromise?

>  I find it good that there’s this MD example in this doc, but I might prefer to have this new part executable without importing other classes (I presume it’s possible, right?)

It's great that you pushed on this! This docstring was actually out of date as (a) now the node libraries are available directly from the workflow, and (b) the default `Node` behaviour has changed to not run automatically, so the MD node needs some tweaking. Just pushing an update now.

We could also consider moving the "output" example further up so it doesn't depend on the "MD" example, which I like, but then we'd need an example higher up with sufficient complexity to have output worth looking at, which I dislike. At the end of the day I'm inclined to leave it as an addendum to MD, but if you feel strongly then I'm willing to slide it up.

samwaseda:
> Hmm, I like the sentiment. I do think it's fair to import `Node` though, since that's the object that gets added to workflows. It _is_ available directly on the workflow now (`wf.add.Node`), but I still think it's nice to see a wild node get added. Does that feel like a fair compromise?

Yeah given the close connection between `Node` and `Workflow`, it's justified to use it.

> We could also consider moving the "output" example further up so it doesn't depend on the "MD" example, which I like, but then we'd need an example higher up with sufficient complexity to have output worth looking at, which I dislike. At the end of the day I'm inclined to leave it as an addendum to MD, but if you feel strongly then I'm willing to slide it up.

That would be perfect!

liamhuber:
> That would be perfect!

I'm happier with it now, good suggestions! Lmk if there's any more you think it needs.

PR #693, opened 2023-05-26
add if statement for ready

### LINKED ISSUE #688: Should every node run to get the output in workflow?
Let's take this example:

```python
@node("y")
def add_one(x: int) -> int:
    return x + 1

@node("y")
def add_two(x: int) -> int:
    return x + 2
```

The workflow of these two works like this:

```python
n_1 = add_one(x=3)
n_1.run()
n_2 = add_two(x=n_1.outputs.y)
n_2.run()
```

but I would have expected the following code should be also valid (and it isn't):

```python
n_1 = add_one(x=3)
n_2 = add_two(x=n_1.outputs.y)
n_2.run()
```

Is this supposed to be like this?

samwaseda:
Closes #688

### DIFF
--- a/pyiron_contrib/workflow/node.py
+++ b/pyiron_contrib/workflow/node.py
@@ -472,6 +472,8 @@ class Node(HasToDict):
             self.run()
 
     def run(self) -> None:
+        if not self.ready:
+            raise ValueError(f"{self.label} is not ready")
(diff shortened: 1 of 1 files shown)


liamhuber:
This is just `Node.update`.

PR #700, opened 2023-05-31
Fitsnap update

mgt16-LANL:
Reduced fitsnap job and removed some dependencies moved to fitsnap.


mgt16-LANL:
Updated and now has ability to pass generated potential information directly to subsequent lammps/pyiron runs!

pmrv:
Thanks for your contribution!  No issues from my side with the new code, but it would be nice if you could additionally integrate fitsnap with the [existing interface for machine learning potentials](https://github.com/pyiron/pyiron_contrib/blob/main/pyiron_contrib/atomistics/ml/potentialfit.py).    This would mostly mean renaming `get_potential` to `get_lammps_potential` and allowing to input `TrainingContainers`.  If the latter is too much work right now, doing it in another PR is fine, though.

pmrv:
Thanks, fair enough!  It's ok to keep in on the radar and not accidentally build a inconsistent interface.

pmrv:
Closing because of inactivity.  If you want to carry this over to [potentialfit](https://github.com/pyiron/pyiron_potentialfit), create a patch file with `git diff` or similar, change the paths (`sed -e /contrib\/atomistics/potentialfit` likely enough) and apply the patch over there with `git apply`.

Issue #677, opened 2023-05-12
Tinybase: Basic.ipynb throwing an error


liamhuber:
I started playing around with `concurrent.futures` with the intent of adding a simple executor class using `ProcessPoolExecutor` to `pyiron_contrib.workflow` in order to bring it closer to `pyiron_contrib.tinybase`.

I quickly learned that `ProcessPoolExecutor` does not play nicely with the interactive interpreter (i.e. my jupyter notebook), because it needs to be able to import its tasks from `__main__` ([docs](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor)). E.g., I have a little `scratch.py` file with the content

```python
def return_one():
    return 1
```

Then, in my notebook I get the following:

```python
def return_two():
    return 2

from scratch import return_one

from concurrent.futures import ProcessPoolExecutor

ex = ProcessPoolExecutor()
f = ex.submit(return_one)
print(f.done(), f.result()) # (True, 1)

# f = ex.submit(return_two)  # BrokenProcessPool error
```

So far so good, this is just what I expect from the docs. But, of course, I want users to be able to define new nodes in their notebook and be able to run them in the background of their workflow! So I need some sort of solution. 

This led me to look directly at `tinybase`. There is significantly more misdirection than in that simple example above, and I wondered if perhaps somehow the demo was working because the task is somehow being imported from a class defined in the (importable!) source code, and the notebook-defined functionality is being attached to that class and "appearing" as importable or something like this...

But actually, when I try to use `tinybase` instead, I get the exact same error!

```python
def return_two():
    return 2

from scratch import return_one

from pyiron_contrib.tinybase.task import FunctionTask
from pyiron_contrib.tinybase.executor import ProcessExecutor

exe = ProcessExecutor(max_processes=1).submit([FunctionTask(return_one)])
exe.run()
exe.wait()
print(exe.output[0].result)  # 1

exe = ProcessExecutor(max_processes=1).submit([FunctionTask(return_two)])
exe.run()  # BrokenProcessPool error
```

I thought maybe this was because I'm defining a function with no input, or because there was some deep and lucky magic in the notebook history that made the `ProcessExecutor` example work, so I just copied `notebooks/tinybase/Basic.ipynb` to a new working directory and ran it directly... and am getting _exactly_ the same `BrokenProcessPool` error for `ProcessExecutor` there too.

[...]

PR #707, opened 2023-06-05
Prepare for executors

liamhuber:
Or servers, or task managers, or whatever you want to call them. The point is that the current implementation simply executes nodes using the main python process, but we want to make space for the node to ship the node functionality (callable and arguments) off to some other entity that will handle the actual computation.

@samwaseda, I think for you're demo I think you'll want to do [something like this](https://stackoverflow.com/questions/2231227/python-subprocess-popen-with-a-modified-environment). I don't know exactly how it will pan out, but be aware that `concurrent.futures.ProcessPoolExecutor` has trouble when combined with the interactive interpreter (i.e. notebooks) on everything _except_ linux, cf. #677. So depending how you go here you might bump into that. Getting a not-just-the-same-process executor/server/task manager running is also relatively high priority for me, so I'll try to be highly responsive to your next step(s).

### DIFF
--- a/pyiron_contrib/workflow/node.py
+++ b/pyiron_contrib/workflow/node.py
@@ -485,6 +485,18 @@ class Node(HasToDict):
             self.failed = True
             raise e
 
+        self.process_output(function_output)
+
+    def process_output(self, function_output):
+        """
+        Take the results of the node function, and use them to update the node.
+
+        By extracting this as a separate method, we allow the node to pass the actual
+        execution off to another entity and release the python process to do other
+        things. In such a case, this function should be registered as a callback
+        so that the node can finishing "running" and push its data forward when that
+        execution is finished.
+        """
         if len(self.outputs) == 1:
             function_output = (function_output,)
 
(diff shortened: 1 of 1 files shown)


samwaseda (approved):

samwaseda:
> @samwaseda, I think for you're demo I think you'll want to do [something like this](https://stackoverflow.com/questions/2231227/python-subprocess-popen-with-a-modified-environment). I don't know exactly how it will pan out, but be aware that `concurrent.futures.ProcessPoolExecutor` has trouble when combined with the interactive interpreter (i.e. notebooks) on everything _except_ linux, cf. #677. So depending how you go here you might bump into that. Getting a not-just-the-same-process executor/server/task manager running is also relatively high priority for me, so I'll try to be highly responsive to your next step(s).

You mean in order to set a different environment? I'm not sure if it works for python though.

samwaseda:
Ah wait `os.environ` is the python environment? Then it might be exactly what we were looking for.

liamhuber:
Yeah, I'm just not sure whether `subprocess.Popen` will face any of the same issues with the interactive interpreter is all

PR #715, opened 2023-06-13
Add calc static

liamhuber:
New node. Maybe many output components should be parsed to not be wrapped in a list, since calc static _always_ returns a single frame?


liamhuber:
@samwaseda, I like that this spawned #717, but since it's just "content" I'll merge it. Then we can apply the ideas of #717 in a later PR.

samwaseda:
> @samwaseda, I like that this spawned #717, but since it's just "content" I'll merge it. Then we can apply the ideas of #717 in a later PR.

Fine by me

samwaseda (approved):

PR #716, opened 2023-06-13
Workflow: Introduce meta nodes

liamhuber:
I.e. a set of callables that return a node _class_ instead of a node _instance_, where the class follows a particular pattern. This is roughly how I envision a lot of flow control happening, but it's just a quick and dirty first pass since I wanted to shorted up an example (now at the end of the examples notebook) for a talk at UBC tomorrow -- i.e. I am not married to this PR and wouldn't be surprised if we close un-merged.

However, I thought it would be a good and practical jumping-off point for discussing meta nodes. Do your worst, @samwaseda 😂 

## EDIT 2023-08-11:

This is still _super_ rough, but as of writing you can do things this:

```python
import numpy as np

n = 5

bulk_loop = Workflow.create.meta.for_loop(
    n,
    Workflow.create.atomistics.Bulk,
    iterate_on = ("a",),
)()

out = bulk_loop(
    name="Al",  # Sent equally to each body node
    A=np.linspace(3.9, 4.1, n).tolist(),  # Distributed across body nodes
)

[struct.cell.volume for struct in out.STRUCTURE]
>>> [14.829749999999995,
...  15.407468749999998,
...  15.999999999999998,
...  16.60753125,
...  17.230249999999995]
```

## EDIT 2023-08-24:

Before going on vacation I thought there was a bug when combining this with macros, but it looks like it was just user error (I was using the wrong variable name and it was trying to iterate over `NotData`). The example below just uses a quadratic fit and not a proper Murnaghan equation, but is otherwise equivalent to a bulk modulus calculation:

```python
import numpy as np
from scipy.optimize import curve_fit

# Define tools 

@Workflow.wrap_as.macro_node()
def lammps_static(macro):
    """
    Species and lattice constant in, 
    cubic structure and Lammps-evaluated energy out.
    """
    macro.structure = macro.create.atomistics.Bulk(
        run_on_updates=True,
        channels_requiring_update_after_run=["name", "a"],
        # Still insufficient flow control when putting a macro in a for loop
        # runs waaaaay more times than necessary
        cubic=True,
    )
    macro.engine = macro.create.atomistics.Lammps(
        structure=macro.structure, 
        run_on_updates=False
    )
    macro.calc = macro.create.atomistics.CalcStatic(job=macro.engine)
    
    macro.engine.signals.input.run = macro.structure.signals.output.ran
    macro.calc.signals.input.run = macro.engine.signals.output.ran
    
    macro.inputs_map = {
        "structure__name": "species", 
        "structure__a": "a",
    }
    macro.outputs_map = {
        "calc__energy_pot": "energy",
        "structure__structure": "structure",
    }
    
def parabola(x, a, b):
    return a * x**2 + b
    
@Workflow.wrap_as.single_value_node()
def ev(structures, energies):
    """
    We should do real Murnaghan, but here's a simple quadratic fit
    """
    params = curve_fit(
        parabola, 
        [struct.cell.volume for struct in structures], 
        np.array(energies).flatten()
    )[0]
    return params

# Build the actual workflow

## Make a properly parameterized for-loop node
n_volumes = 5

BulkLoop = Workflow.create.meta.for_loop(
    lammps_static,
    n_volumes,
    iterate_on="a",
)

## Build the workflow
wf = Workflow("my_ev")

wf.bulk_loop = BulkLoop(
    species="Al",  # Sent equally to each body node
    A=np.linspace(3.9, 4.1, n_volumes).tolist(),  # Distributed across body nodes
)

wf.ev_curve = ev(
    structures=wf.bulk_loop.outputs.STRUCTURE, 
    energies=wf.bulk_loop.outputs.ENERGY
)

wf.outputs_map = {"ev_curve__params": "params"}

wf.outputs.params.value
>>> array([-6.45915414e-04, -3.17537394e+00])
```

So everything is working so far, but I want to add a `while` loop and some tests before turning off "draft" here.

## EDIT 2023-08-25

The while-loop was actually comparatively simple after the for loop, which surprised me a little. You can now do things like this:

```python
import numpy as np
np.random.seed(0)

# Build tools

@Workflow.wrap_as.single_value_node()
def random(length: int | None = None):
    random = np.random.random(length)
    return random

@Workflow.wrap_as.single_value_node()
def greater_than(x: float, threshold: float):
    gt = x > threshold
    symbol = ">" if gt else "<="
    print(f"{x:.3f} {symbol} {threshold}")
    return gt

RandomWhile = Workflow.create.meta.while_loop(random)

# Define workflow

wf = Workflow("random_until_small_enough")

## Wire together the while loop and its condition

wf.gt = greater_than()
wf.random_while = RandomWhile(condition=wf.gt)
wf.gt.inputs.x = wf.random_while.Random

wf.starting_nodes = [wf.random_while]

## Give convenient labels
wf.inputs_map = {"gt__threshold": "threshold"}
wf.outputs_map = {"random_while__Random__random": "capped_value"}

# Set a threshold and run

print(f"Finally {wf(threshold=0.1).capped_value:.3f}")
>>> 0.549 > 0.1
>>> 0.715 > 0.1
>>> 0.603 > 0.1
>>> 0.545 > 0.1
>>> 0.424 > 0.1
>>> 0.646 > 0.1
>>> 0.438 > 0.1
>>> 0.892 > 0.1
>>> 0.964 > 0.1
>>> 0.383 > 0.1
>>> 0.792 > 0.1
>>> 0.529 > 0.1
>>> 0.568 > 0.1
>>> 0.926 > 0.1
>>> 0.071 <= 0.1
>>> Finally 0.071
```

@samwaseda, IMO this is ready for some feedback now. Because I'm anticipating significant changes to the interface, and indeed even what sorts of promises we want to be making here, I didn't add exhaustive tests. However, I _did_ add these examples to the integration tests, so we'll at least get some heads-up if something breaks the basic premise.

Before unit tests and notebook examples, I want to have an earnest go at converting from a push-based to a pull-based paradigm to see if this can simplify a lot of the stuff we currently do with `Signals`. Having the while- and for-loop in place should make it easier to test and appreciate pull-paradigm changes.

## EDIT 2023-08-29

No changes to the code, just another example. Here is one with looped-data with initialization (since setting the initial value doesn't destroy the existing connection). It is analogous to the [example in jobflow](https://materialsproject.github.io/jobflow/tutorials/5-dynamic-flows.html), but (+) the decorated functions don't need to know about our classes, (+, IMO) it re-uses the body node instead of making a million copies, and (-) it's way more work to actually set up the workflow. I find the contrast useful for setting goals for simplicity here.

```python

@Workflow.wrap_as.single_value_node(run_on_updates=False)
def add(a, b):
    return a + b

@Workflow.wrap_as.single_value_node()
def less_than_ten(value):
    return value < 10

AddWhile = Workflow.create.meta.while_loop(add)

wf = Workflow("do_while")
wf.lt10 = less_than_ten()
wf.add_while = AddWhile(condition = wf.lt10)

wf.lt10.inputs.value = wf.add_while.Add
wf.add_while.Add.inputs.a = wf.add_while.Add

wf.starting_nodes = [wf.add_while]
wf.inputs_map = {
    "add_while__Add__a": "a", 
    "add_while__Add__b": "b"
}
wf.outputs_map = {"add_while__Add__a + b": "total"}

out = wf(a=1, b=2)
print(out.total)
>>> 11
```


liamhuber:
I expect significant change after discussion, but even if that doesn't happen this still needs tests before it's not WIP.

liamhuber:
So, actually I guess this is more of a "dynamic" node than a "meta" node?

liamhuber:
Notebook failures is just a [known issue](https://github.com/pyiron/pyiron_contrib/pull/824#issuecomment-1692211906) with the Tinybase notebooks and not related to this PR.

liamhuber:
Ok, I'm pretty happy with this. The interface leaves something to be desired still, so it's in the example notebook under a "Here be dragons" section where we warn of possible sudden and drastic changes. The third example where while-looped data works fine was the deal-maker for me.

I plan to merge this and move on with other PRs. I expect substantial changes to the UI here if pull-mode winds up working.

PR #718, opened 2023-06-16
Add a Dask Executor

pmrv:
I've only tested this with the local dask cluster, because I haven't yet managed to setup an SSH cluster (for lack of remote machines).  However there's no fundamental reason why this shouldn't work, so this opens up the route to working with proper HPC queues (via dask_jobqueue) and remote setups as we know them from base with very little effort from us.


jan-janssen:
I guess the limitation of DASK is that they only support serial python tasks, in their scheduling, that is why in the Exascale workshop we first looked at parsl, which at least supports calling external executables  https://github.com/Parsl/parsl/issues/2723 and finally settled on https://flux-framework.org . Based on the combination of https://github.com/pyiron/pympipool and flux we cover all four categories of tasks: 

| **Code type / Parallelism**             | **serial**        | **MPI parallel** |
|-----------------------------------|---------------|-----------------|
| **Python task**                                | pympipool      | pympipool           | 
| **Code without Python bindings** | flux                 | flux                       |

With `pympipool` using `flux` as a backend. So I feel we have the task management side covered, the missing pieces are now the storage interface and the integration with the graph based workflows. Just like we discussed two weeks ago in the pyiron meeting: https://github.com/pyiron/team/discussions/176

pmrv:
Agree that flux seems nicer.  I had actually started this morning trying to add that instead, but then noticed that it doesn't support submitting arbitrary callables, but only executables.   That's not a deal breaker, but requires a little more code from me.  Since `pympipool` offers already an `Executor`  interface, it should be similarly simple to integrate as dask.

jan-janssen:
Yes, that is exactly the purpose of `pympipool` https://pympipool.readthedocs.io/en/latest/interfaces.html

If you use the `Executor` interface you can even run MPI4py parallel functions:
```
from pympipool import Executor

def calc(i):
    from mpi4py import MPI
    size = MPI.COMM_WORLD.Get_size()
    rank = MPI.COMM_WORLD.Get_rank()
    return i, size, rank

with Executor(cores=2, enable_flux_backend=True) as p:
    fs = p.submit(calc, 3)
    print(fs.result())
```
The optional parameter `enable_flux_backend=True` switches the backend from MPI to flux. But the limitation of this implementation is that it only executes one function at a time. 

If you want something more like dask you can use:
```
from pympipool import PoolExecutor

def calc(i, j):
    return i + j

with PoolExecutor(max_workers=2, enable_flux_backend=True) as p:
    fs1 = p.submit(calc, 1, 2)
    fs2 = p.submit(calc, 3, 4)
    fs3 = p.submit(calc, 5, 6)
    fs4 = p.submit(calc, 7, 8)
    print(fs1.result(), fs2.result(), fs3.result(), fs4.result())
```
The limitation of this interface is that it only supports serial python functions. The `mpi4py` developers currently extend the support for parallel tasks so I am optimistic that the `PoolExecutor` interface gets support for parallel functions soon: 
https://github.com/mpi4py/mpi4py/pull/369

jan-janssen:
Yes, `pympipool` by default only requires `mpi4py` which is platform independent. The flux integration is only important when we have all four categories of tasks: serial and parallel python tasks in addition to serial and parallel tasks without python bindings. For the later part we use flux and for the python based tasks pympipool with the flux backend. This way all resources are tracked by flux but a subset of the resources is made accessible to python via pympipool.

liamhuber:
> Yes, `pympipool` by default only requires `mpi4py` which is platform independent. The flux integration is only important when we have all four categories of tasks: serial and parallel python tasks in addition to serial and parallel tasks without python bindings. For the later part we use flux and for the python based tasks pympipool with the flux backend. This way all resources are tracked by flux but a subset of the resources is made accessible to python via pympipool.

Ok super, then it sounds to me like in principle one could be accomplishing a similar task to this PR using `pympipool`, but in principle there is no fundamental problem with also having a dask executor as implemented in this existing PR.

jan-janssen:
> Ok super, then it sounds to me like in principle one could be accomplishing a similar task to this PR using `pympipool`, but in principle there is no fundamental problem with also having a dask executor as implemented in this existing PR.

I am fine with having the Executor be a variable the user can set, as different users might have different preferences. Still I think pyiron should provide a suggestion which Executor is recommended for which setting and I am not sure if I would recommend task in this perspective.

pmrv:
Very sure that we want to have the exectutor as a user choice (as we have the run mode just now).  In my head canon there would be a config file that maps human readable names to these executor classes and their arguments, but that's still a bit out.  I would like to postpone the discussion of defaults until we've had actual experience running real calculations, but I have no strong preference either way.

PR #724, opened 2023-06-19
Add server object to Node and Workflow

samwaseda:


liamhuber:
This looks fine to me. We'll need to make sure server settings get wrapped into the serialization, but server stuff is "meta" and doesn't belong in the IO, so this is all fine as-is.

PR #730, opened 2023-06-21
Add pympipool executor

pmrv:
Adds also the dependency on distributed, because I forgot that over in #718 .

### DIFF
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -20,5 +20,7 @@ dependencies:
 - pycp2k =0.2.2
 - typeguard =4.0.0
 - aws-sam-translator =1.68.0
+- pympipool =0.5.0
+- distributed =2023.6.0
 - python >= 3.10
 - lammps
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -21,6 +21,6 @@ dependencies:
 - typeguard =4.0.0
 - aws-sam-translator =1.68.0
 - pympipool =0.5.1
-- distributed =2023.6.0
+- distributed =2023.5.0
 - python >= 3.10
 - lammps
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -20,7 +20,7 @@ dependencies:
 - pycp2k =0.2.2
 - typeguard =4.0.0
 - aws-sam-translator =1.68.0
-- pympipool =0.5.0
+- pympipool =0.5.1
 - distributed =2023.6.0
 - python >= 3.10
 - lammps
--- a/.ci_support/environment.yml
+++ b/.ci_support/environment.yml
@@ -20,5 +20,5 @@ dependencies:
 - pycp2k =0.2.2
 - typeguard =4.0.0
 - aws-sam-translator =1.68.0
-- pympipool =0.5.0
+- pympipool =0.5.1
 - distributed =2023.6.0
--- a/.ci_support/environment.yml
+++ b/.ci_support/environment.yml
@@ -20,3 +20,5 @@ dependencies:
 - pycp2k =0.2.2
 - typeguard =4.0.0
 - aws-sam-translator =1.68.0
+- pympipool =0.5.0
+- distributed =2023.6.0
--- a/.ci_support/environment.yml
+++ b/.ci_support/environment.yml
@@ -21,4 +21,4 @@ dependencies:
 - typeguard =4.0.0
 - aws-sam-translator =1.68.0
 - pympipool =0.5.1
-- distributed =2023.6.0
+- distributed =2023.5.0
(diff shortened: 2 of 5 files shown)


.binder/environment.yml
```diff
@@ -20,5 +20,7 @@ dependencies:
 - pycp2k =0.2.2
 - typeguard =4.0.0
 - aws-sam-translator =1.68.0
+- pympipool =0.5.0
```
jan-janssen, line 23:
```suggestion
- pympipool =0.5.1
```

.ci_support/environment.yml
```diff
@@ -20,3 +20,5 @@ dependencies:
 - pycp2k =0.2.2
 - typeguard =4.0.0
 - aws-sam-translator =1.68.0
+- pympipool =0.5.0
```
jan-janssen, line 23:
```suggestion
- pympipool =0.5.1
```

docs/environment.yml
```diff
@@ -22,3 +22,5 @@ dependencies:
 - pycp2k =0.2.2
 - typeguard =4.0.0
 - aws-sam-translator =1.68.0
+- pympipool =0.5.0
```
jan-janssen, line 25:
```suggestion
- pympipool =0.5.1
```

setup.py
```diff
@@ -57,6 +57,10 @@
             'python>=3.10',
             'ipython',
             'typeguard==4.0.0'
+        ],
+        'tinybase': [
+            'distributed==2023.6.0',
+            'pympipool==0.5.0'
```
jan-janssen, line 63:
```suggestion
            'pympipool==0.5.1'
```

jan-janssen:
With the new release of `pympipool=0.5.1` this should work fine.

.binder/environment.yml
```diff
@@ -20,5 +20,7 @@ dependencies:
 - pycp2k =0.2.2
 - typeguard =4.0.0
 - aws-sam-translator =1.68.0
+- pympipool =0.5.1
+- distributed =2023.6.0
```
jan-janssen, line 24: @pmrv Is `distributed` required? It seems to be causing the tests to crash and `pympipool` does not require it.

liamhuber: Isn't it the `distributed` package that [provides the `Futures` interface to dask](https://distributed.dask.org/en/stable/)? If so, I imagine (but am not certain as I'm not tracking this super closely) that this is required to exploit dask in the futures-based paradigm.

I've only played with a few of the basic examples on the dask page, so I definitely could be missing something. But, at a minimum, dropping `distributed` would require re-implementing [the existing dask executor](https://github.com/pyiron/pyiron_contrib/blob/5064692b9bed95a313f9615a8ce748721a39ca34/pyiron_contrib/tinybase/executor.py#L257)

jan-janssen: Oh I see `from dask.distributed import Client, LocalCluster` dask requires `distributed` while in `pympipool` we use the `concurrent.futures` from python so there we do not need it.

jan-janssen: It is one more python 3.8 issue. `distributed=2023.5.0` still supports python 3.8 while `distributed=2023.6.0` drops support for 3.8. So I would be in favor of removing 3.8.

liamhuber: Yeah, it's super easy to miss in the code because it _really_ looks like a submodule, but on conda it gets its own unique tag.

It's possible there's some alternative for the `DaskExecutor` that would only lean on `concurrent.futures` and the older parts of dask, but I would definitely prefer to first attempt a direct resolution to the failing tests so we can keep the more modern dependency and play with all of dask's bells and whistles.

liamhuber: > It is one more python 3.8 issue. `distributed=2023.5.0` still supports python 3.8 while `distributed=2023.6.0` drops support for 3.8. So I would be in favor of removing 3.8.

Agreed! I've got an [open PR](https://github.com/pyiron/actions/pull/46) primed and ready to go that will do this, just waiting on the MPIE cluster to be ready.

That's good news that "fixing" the tests will be so simple!

jan-janssen: As long as we stay on one node the `concurrent.futures.Executor` class offers all we need and when we move beyond one node and stay in an HPC environment I would suggest using `pympipool` with `mpi` backend and only for very large allocations when we want python tasks and external executables next to each other then I would move to `flux`. Again `flux` comes with the restriction that it only works on Linux.

pmrv: For now I have simply dropped the version of `distributed`, once [https://github.com/pyiron/actions/pull/46] is in I suppose dependabot will take care of it.

jan-janssen (approved):
Looks good to me

PR #731, opened 2023-06-21
Use the new workflow that swaps 3.8 for 3.11

liamhuber:
I'm sure there must be a way to automate this sort of test, but for this exact moment let's just be quick and dirty and manually swap the branch target in a PR on this repo. There is no intention to actually merge this commit to main.

### DIFF
--- a/.github/workflows/push-pull-main.yml
+++ b/.github/workflows/push-pull-main.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   pyiron:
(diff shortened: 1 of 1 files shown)


liamhuber:
Ok, the 3.8 tests disappear and 3.11 ran perfectly. 3.10 remains the "main" test version (for windows, mac, and non-unit test jobs), but IMO that is ok since other non-pyiron packages have not upgraded. We can come in and make 3.11 the "main" version in another PR at any time in the future.

PR #733, opened 2023-06-22
Workflow: Merge changes I made so far

samwaseda:
I open this one now because otherwise this branch only becomes larger and deviates from `main`.


liamhuber:
> I open this one now because otherwise this branch only becomes larger and deviates from `main`.

Super. I was thinking this too! I lean towards merging this before merging #729, and I'll handle the merge to rebase this stuff onto the new abstract class.

liamhuber:
I'll hold off a full review until we get the stacked PRs merged in here?

Taking a quick glance, one thing I think would be really nice is having a concrete example in `notebooks/workflow_example.ipynb` of the new node-with-self functionality being used. If I think of a good one today I'll post it here.

samwaseda:
> I'll hold off a full review until we get the stacked PRs merged in here?

Is there a particular reason? I have the feeling that they make this PR only larger

> Taking a quick glance, one thing I think would be really nice is having a concrete example in `notebooks/workflow_example.ipynb` of the new node-with-self functionality being used. If I think of a good one today I'll post it here.

So far there's nothing that really changes the functionality. As soon as there's a possibility to set it up (especially the server), I would say we can add it.

liamhuber:
>> I'll hold off a full review until we get the stacked PRs merged in here?
> Is there a particular reason? I have the feeling that they make this PR only larger

The reason had been that tests were failing and I thought this _needed_ the other stuff. It looks like the only other connected PR is #725. I'll just review each separately, but I think we can merge both soon.

> So far there's nothing that really changes the functionality. As soon as there's a possibility to set it up (especially the server), I would say we can add it.

The ability to include `self` is definitely a functionality change though, and an example can be included independently. Probably in the easiest case it will need to be a sort of "counter example" in that using `self` will violate the "functional" aspect of the node by introducing some state -- e.g. we could make a node with an internal counter that keeps track of how many times it's been called. There should be _some_ example of the new functionality though.

Regarding `Server`, I guess it will be hard to make a solid example there, because queues need to be set up locally... What do you think of unwinding the `Server` changes in this PR and having it be _just_ the `self` stuff? Then we could make a separate PR where we introduce `Executor` instead of `Server` and exploit Marvin's Dask executor form #718 to build an architecture-independent example of running a node task in the background? Look at the diff, unwinding the server stuff would only involve removing the import and `self.server=` assignment in node.py and workflow.py, so it's a pretty easy change.

pyiron_contrib/workflow/node.py
```diff
@@ -14,6 +14,7 @@
 from pyiron_contrib.workflow.has_channel import HasChannel
 from pyiron_contrib.workflow.has_to_dict import HasToDict
 from pyiron_contrib.workflow.io import Inputs, Outputs, Signals
+from pyiron_base.jobs.job.extension.server.generic import Server
```
liamhuber, line 17:
```suggestion
```

Roll back server changes

pyiron_contrib/workflow/node.py
```diff
@@ -352,16 +353,30 @@ def __init__(
             elif k not in self._init_keywords:
                 warnings.warn(f"The keyword '{k}' was received but not used.")
         self.run_on_updates = run_on_updates
+        self.server = Server()
```
liamhuber, line 356:
```suggestion
```

Roll back server changes

pyiron_contrib/workflow/workflow.py
```diff
@@ -4,6 +4,7 @@
 from pyiron_contrib.workflow.has_to_dict import HasToDict
 from pyiron_contrib.workflow.node import Node, node, fast_node, single_value_node
 from pyiron_contrib.workflow.util import DotDict
+from pyiron_base.jobs.job.extension.server.generic import Server
```
liamhuber, line 7:
```suggestion
```

Roll back server changes

pyiron_contrib/workflow/workflow.py
```diff
@@ -121,6 +122,7 @@ class Workflow(HasToDict, HasNodes):
     def __init__(self, label: str, *nodes: Node, strict_naming=True):
         super().__init__(strict_naming=strict_naming)
         self.label = label
+        self.server = Server()
```
liamhuber, line 125:
```suggestion
```

Roll back server changes

tests/unit/workflow/test_node.py
```diff
@@ -156,18 +157,35 @@ def test_statuses(self):
         # self.assertFalse(n.running)
         self.assertFalse(n.failed, msg="Re-running should reset failed status")
 
-
-@skipUnless(version_info[0] == 3 and version_info[1] >= 10, "Only supported for 3.10+")
-class TestFastNode(TestCase):
+    def test_with_self(self):
+        def with_self(self, x: float) -> float:
```
liamhuber, line 161: For an example in the ipynb, we could even simply modify this with the line `self.all_my_exes = [x] if self.all_my_exes is None else self.all_my_exes + [x]` to make a (trivial) example of how `self` can be exploited in the function call. It should come with some text saying you probably _shouldn't_ do it this way because it violates the "functional" best-practice, but I just really want some public facing example of how to exploit this new functionality.

liamhuber (changes requested):
Ok, I basically said everything I had to say in the non-review-comment five minutes ago 😝 Here I just act on it.

Purge `server` and add example to the ipynb at it's good-to-go.

samwaseda:
> The reason had been that tests were failing and I thought this _needed_ the other stuff. It looks like the only other connected PR is #725. I'll just review each separately, but I think we can merge both soon.

Ah yeah and the reason why it was failing was because the tests are apparently not triggered when the PR is not to be merged to `main`. There's certainly a way to make them run, but at least for now that's yet another reason why I would like to merge directly to `main`.

> The ability to include `self` is definitely a functionality change though, and an example can be included independently. Probably in the easiest case it will need to be a sort of "counter example" in that using `self` will violate the "functional" aspect of the node by introducing some state -- e.g. we could make a node with an internal counter that keeps track of how many times it's been called. There should be _some_ example of the new functionality though.

Hmm on the other hand there are only counter examples and nothing that enhances the capabilities. Since it's relatively obvious that `self` should not be used, I guess we can wait until further functionalities are added??

> 
> Regarding `Server`, I guess it will be hard to make a solid example there, because queues need to be set up locally... What do you think of unwinding the `Server` changes in this PR and having it be _just_ the `self` stuff? Then we could make a separate PR where we introduce `Executor` instead of `Server` and exploit Marvin's Dask executor form #718 to build an architecture-independent example of running a node task in the background? Look at the diff, unwinding the server stuff would only involve removing the import and `self.server=` assignment in node.py and workflow.py, so it's a pretty easy change.

Fine by me.

liamhuber:
> Hmm on the other hand there are only counter examples and nothing that enhances the capabilities. Since it's relatively obvious that `self` should not be used, I guess we can wait until further functionalities are added??

Mmmm, I hear you.... but right now I'm really gearing `example_workflow.ipynb` towards developers rather than users, i.e. so people like Marvin and Joerg have  a single point of entry where they can come familiarize themselves with all the functionality. From this perspective I feel a counter-example is actually still quite productive.

I can see you might be right and this isn't a hill I need to die on, so If you still disagree with me then let's just drop it and merge sans example, but I hope I was sufficiently persuasive in the last paragraph  😂

samwaseda:
> Mmmm, I hear you.... but right now I'm really gearing `example_workflow.ipynb` towards developers rather than users, i.e. so people like Marvin and Joerg have a single point of entry where they can come familiarize themselves with all the functionality. From this perspective I feel a counter-example is actually still quite productive.

Ah in that case I would rather add comments inside the code, because the code itself is a lot more visible to someone like @pmrv.

liamhuber:
> Ah in that case I would rather add comments inside the code, because the code itself is a lot more visible to someone like @pmrv.

Good point, that is an even better solution.

samwaseda:
Now I updated the docs. I guess it's ready now.

liamhuber:
> Now I updated the docs. I guess it's ready now.

Can we modify `test_with_self` so that it actually assigns something to `self` and then verify after the `run` that this is present? I believe this will work fine, I would just sleep better if there's a test; e.g. to prevent changes to `Node.__setattr__` from quietly breaking our expectations of what users can do by adding `self`.

liamhuber (approved):
I'd optimally still like to see `test_with_self` be more thorough as described [here](https://github.com/pyiron/pyiron_contrib/pull/733#issuecomment-1607902736), but don't want my old "request changes" to block merging.

samwaseda:
Do you have an idea how to use `self` twice? Essentially having the `self` from the function and `self` from the unit tests?

liamhuber:
> Do you have an idea how to use `self` twice? Essentially having the `self` from the function and `self` from the unit tests?

It gets locally scoped, so it's a non-issue.

I thought you might have (very fairly!) stopped for the day, so I implemented what I had in mind in #742. Lmk what you think!

PR #734, opened 2023-06-22
Tinybase: Add creators for UX

pmrv:
Add creator classes for jobs, tasks, structures and executors so that users don't need to import them.  With it there are some changes in TinyJob:

1. it is no longer an abstract class, but requires a task instance to be instantiated.  Reloading already run jobs can be done with TinyJob.restore
2. as a consequence GenericTinyJob is no longer needed
3. because Executors are easily created now, the _executors class attribute is also gone; TinyJob.run now takes an argument 'executor', either an executor or a method name of the ExecutorCreator as a short hand

The creators are very similar to the ones already existing in pyiron_base.  Notable is the ExecutorCreator which as a method `most_recent` that simply returns the last created executor, which is also used by TinyJob.run by default, freeing the user from lugging around executor instances on their own.

### DIFF
--- a/pyiron_contrib/tinybase/creator.py
+++ b/pyiron_contrib/tinybase/creator.py
@@ -90,7 +90,7 @@ class JobCreator(Creator):
                 else:
                     if not isinstance(job.task, task):
                         raise ValueError(
-                            f"Job with given name already exists, but is of different type!"
+                            "Job with given name already exists, but is of different type!"
                         )
                     return job
             except JobNotFoundError:
--- a/pyiron_contrib/tinybase/creator.py
+++ b/pyiron_contrib/tinybase/creator.py
@@ -1,0 +1,250 @@
+"""
+Creators of various things.
+
+They serve to give easy access to various classes, so that users do not have to
+manually import them.
+
+Rough spec:
+
+    1. RootCreator is merely the entry point and delegates to sub creators
+    based on name
+    2. Sub creators should be type homogenous or related by functionality, i.e.
+    one of jobs, one for tasks, one for structures, etc.
+    3. Each project gets its own RootCreator, but should keep that instance
+    alive so that sub creators may cache certain things
+"""
+
+import abc
+import importlib
+from typing import Union
+from functools import wraps
+from os import sched_getaffinity
+
+import pyiron_contrib.tinybase.job
+from pyiron_contrib.tinybase.executor import (
+        Executor,
+        ProcessExecutor,
+        BackgroundExecutor,
+        DaskExecutor
+)
+from pyiron_atomistics import ase_to_pyiron, Atoms
+
+import ase.build
+
+class Creator(abc.ABC):
+
+    @abc.abstractmethod
+    def __init__(self, project, config):
+        pass
+
+    #TODO add all methods
+
+def load_class(class_path):
+    module, klass = class_path.rsplit(".", maxsplit=1)
+    try:
+        return getattr(importlib.import_module(module), klass)
+    except ImportError as e:
+        raise ValueError(f"Importing task class '{class_path}' failed: {e.args}!") from None
+
+class JobCreator(Creator):
+
+    def __init__(self, project, job_dict):
+        self._project = project
+        self._jobs = job_dict.copy()
+
+    def __dir__(self):
+        return tuple(self._jobs.keys())
+
+    def __getattr__(self, task_type):
+        if task_type not in self._jobs:
+            raise ValueError(f"Unknown task type {task_type}!")
+        task = self._jobs[task_type]
+        if isinstance(task, str):
+            task = self._jobs[task_type] = load_class(task)
+
+        def create(
+                name: str,
+                delete_existing_job: bool = False,
+                delete_aborted_job: bool = False
+        ):
+            """
+            Create or load a new job.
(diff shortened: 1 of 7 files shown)


pmrv:
Together with this I'm starting to use our `Atoms` instead of the ase one for the structure creator. That lead to a bunch of `to_ase` calls in the notebooks, but that will be fixed in a different PR.

pyiron_contrib/tinybase/creator.py
```diff
@@ -0,0 +1,255 @@
...
+    def __init__(self, project, config):
+        pass
+
+    @wraps(ase.build.bulk)
+    def bulk(self, *args, **kwargs):
+        return ase_to_pyiron(ase.build.bulk(*args, **kwargs))
+
+    @wraps(Atoms)
+    def atoms(self, *args, **kwargs):
+        return Atoms(*args, **kwargs)
```
liamhuber, line 171: What is the advantage to this over simply `atoms = Atoms`? I see that the defined method prevents overwriting the attribute, but I'm not so worried about that and otherwise I don't see any differences in functionality or documentation access.

For `bulk` I see it right away, because you really wrap it in `ase_to_pyiron` first (maybe there should be an addition to `bulk.__doc__` indicating that the return type changes), but here I'm just curious if I'm missing some helpful behaviour

pyiron_contrib/tinybase/creator.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def __init__(self, project, config):
+        self._most_recent = None
+
+    def most_recent(self):
+        if self._most_recent is None:
+            self._most_recent = Executor()
+        return self._most_recent
+
+    def _save(func):
```
liamhuber, line 185: This gets away without a `staticmethod` decorator??

liamhuber: Ah, right, because the `@_save` invocation also doesn't know about `self` or anything, it's just looking in the scope of the class. Cool. The only ugly part about it is that it's really line-order-dependent, so you can't just slide `_save` down below, it _has_ to appear before its invocations. That's not a high price or anything though.

pyiron_contrib/tinybase/creator.py
```diff
@@ -0,0 +1,255 @@
+"""
+Creators of various things.
+
+They serve to give easy access to various classes, so that users do not have to
+manually import them.
+
+Rough spec:
```
liamhuber, line 7: This is great, we should do more of this. I am really making an effort to at least do it at the class level, but it's really nice to see at the module level like this too!!!

pyiron_contrib/tinybase/creator.py
```diff
@@ -0,0 +1,255 @@
...
+            return self._subcreators[name]
+        except KeyError:
+            raise AttributeError(f"No creator of name {name} registered!")
+
+
+TASK_CONFIG = {
+    "AseStatic": "pyiron_contrib.tinybase.ase.AseStaticTask",
+    "AseMinimize": "pyiron_contrib.tinybase.ase.AseMinimizeTask",
+    "AseMD": "pyiron_contrib.tinybase.ase.AseMDTask",
+    "Murnaghan": "pyiron_contrib.tinybase.murn.MurnaghanTask",
```
liamhuber, line 246: I always get nervous about maintainability when I see heavy use of hard-coded strings like this; is this necessary to avoid circular imports?

If not, is it possible to import the relevant objects and back the import path out directly from the objects? Then if they get moved or something in their path gets renamed, `TASK_CONFIG` will stay up-to-date by default. E.g. `from pyiron_contrib.tinybase.ase import AseStaticTask` and `"AseStatic": get_class_string(AseStaticTask)` where `def get_class_string(klass): return str(klass).split("'")[1]` (or whatever is most robust and readable in this function).

liamhuber:
Cool! I didn't do a real review, but left a couple of comments I hope are helpful, and I enjoyed giving it a quick read-through 😄

pyiron_contrib/tinybase/creator.py
pmrv: I didn't actually consider it.  It would have to be a method though, I guess, so 

```python
atoms = staticmethod(Atoms)
```

would work I suppose, but look a bit wonky to me.

pyiron_contrib/tinybase/creator.py
pmrv: Yeah, `_save` is really just a convenience decorator, it could just as well be defined outside the class, probably that's cleaner anyway.

pyiron_contrib/tinybase/creator.py
liamhuber: Is there a reason it needs to be a method? Over in workflows [I just slapped a whole class as an attribute on another class like this](https://github.com/pyiron/pyiron_contrib/blob/9378cc9e362f17fb91ecb9343435175a0f986386/pyiron_contrib/workflow/has_nodes.py#L146) and so far haven't run into any side effects.

pmrv: Fair I hadn't considered that.

pmrv:
Not perfectly happy with the `Creator` "base" class, but I'll merge it for the other changes anyway.

PR #736, opened 2023-06-22
Workflow: Remove storage priority

liamhuber:
It didn't actually do anything yet, so let's hold off implementing it until we have serialization for it to work with.


liamhuber:
I don't hate the idea of having storage priorities in roughly this way, I just feel that I did some premature optimization and want to unwind it.

samwaseda (approved):
Yeah I agree that it's better to not offer something that doesn't exist XD but it's gonna hopefully come back very soon in the future.

PR #738, opened 2023-06-23
 Tinybase: Rework storage interface

pmrv:
Previously Storable implementions of `AbstractTask` and `TinyJob` simply
pickled everything at the highest level because some ASE objects
couldn't be written directly with the Storable interface.  This change
adds some functionality to GenericStorage that will automatically detect
such objects and pickle only those transparently to the rest of
tinybase. As such the full object hierarchy is now represented in
storage and neither tasks nor jobs need to know whether some internal
objects of theirs have been pickled.

It also adds some amount of docstrings and specs to the storage related
classes.

Pending #734


pyiron_contrib/tinybase/storage.py
```diff
@@ -310,3 +383,30 @@ def _restore(cls, storage, version):
...
+
+def pickle_load(buf):
+    return pickle.loads(codecs.decode(buf.encode(), "base64"))
+
+
+class PickleStorable(Storable):
+    """
+    Trivial implementation of :class:`.Storable` that pickles values.
+
+    Used as a fallback by :class:`.ProjectHDFioStorageAdapter` if value cannot
```
pmrv, line 400:
```suggestion
    Used as a fallback by :class:`.GenericStorage` if value cannot
```

pmrv:
@liamhuber This fixes now most of the CI problems with tinybase.  The last error is related to some environment hacks.  I had added a pyiron `resource` folder to the notebooks folder so that the notebook can demonstrate how to run custom executables.  I'm guessing that because the notebook runner runs from a weird working directory that those are not correctly found in the CI, though it works locally.   I'll merge this here anyway, but if we can't find a solution I can just remove the notebook or add it to the exclusion file.

PR #742, opened 2023-06-26
Test self capability

liamhuber:


liamhuber:
We should consider changing the centralized CI so tests run on all PRs... But anyhow, if you like how this is formulated there's no need to check it out on your local machine -- it runs on mine fine, so we can directly merge it to `submittable_workflow` and let the CI double check it across architectures there.

samwaseda:
> We should consider changing the centralized CI so tests run on all PRs... But anyhow, if you like how this is formulated there's no need to check it out on your local machine -- it runs on mine fine, so we can directly merge it to `submittable_workflow` and let the CI double check it across architectures there.

Ah right! That was a point that I wanted to raise today, but I totally forgot.

PR #743, opened 2023-06-27
Workflow: Draft spec

liamhuber:
So the purpose of this is twofold: First, we talked quite a while ago about adding `README.md` files to different contrib sub-modules -- since the submodules are all so different compared to in other pyiron repos -- to see if they get nicely rendered by GitHub; Second, as a place to discuss overall code structure and start defining some interfaces/expectations/promises, even before they're implemented!

@samwaseda, let me know what you think about the content, but also the idea of using this space for these sorts of discussions/definitions, at least while we're devving. Note that the actual list of classes is really only a temporary feature since a bunch of them aren't defined -- ultimately we can replace that with an automatically generated UML diagram and some light commentary where needed.

### DIFF
--- a/pyiron_contrib/workflow/README.md
+++ b/pyiron_contrib/workflow/README.md
@@ -115,8 +115,13 @@ Ultimately, one will be able to automatically extract this information from the 
     - `input` and (not implemented yet) `output` channels automatically created, named, and typed from the function
     - `run_args` automatically generated from the `input`
     - Best practices: node function should be functional and idempotent
-    - `Fast(Node)`
-      - Always 
+    - `Fast(Function)`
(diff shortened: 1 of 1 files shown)


liamhuber:
Ok, rad [GitHub does render it nicely!](https://github.com/pyiron/pyiron_contrib/tree/add_workflow_readme/pyiron_contrib/workflow)

liamhuber:
In the process of making this document and playing in #729  and #745, I realized that there are times that we want to assume that `HasNodes` has a `label` and so forth, and that `Macro` and `Workflow` will probably share implementations of required `Node` (formerly/presently `IsNodal`) methods.... so it makes a lot more sense to interject a common ancestor -- which I call `Composite` here -- instead of using a mix-in strategy with `HasNodes`.

PR #745, opened 2023-06-28
Workflow: Centralize working directory

### LINKED ISSUE #744: Workflow: Why can't `Node` have a `working_directory`
without a parent `Workflow`? 

I can imagine that you make a workflow with some label, run it, and populate a folder. We then don't want to instantiate another workflow or another node with the same label at that level, lest their working directories conflict. Instead of forbidding nodes from having their own working directories, could we instead check for populated working directories at either instantiation or serialization time and raise an error then if the label is already in use?

liamhuber:
Moves the working directory up to the `IsNodal` mix-in class. In the process I had to bring up the `parent` attribute too; `Node.parent` works as before, but here `Node.working_directory` is allowed to exist even without a `parent`. `Workflow.parent` now exists but conforms to the existing spec that workflows can't have parents -- i.e. it's forced to only ever be `None` and still doesn't appear as an `__init__` arg.

If we want to change the node spec this way, this PR closes #744.


pyiron_contrib/workflow/workflow.py
```diff
@@ -178,3 +172,17 @@ def update(self):
     def on_run(self):
         # Maybe we need this if workflows can be used as nodes?
         raise NotImplementedError
+
+    @property
+    def parent(self) -> None:
+        return self._parent
```
samwaseda, line 178:
```suggestion
        return None
```

pyiron_contrib/workflow/workflow.py
```diff
@@ -121,17 +120,12 @@ class Workflow(IsNodal, HasToDict, HasNodes):
     wrap_as = _NodeDecoratorAccess
 
     def __init__(self, label: str, *nodes: Node, strict_naming=True):
+        self._parent = None
```
samwaseda, line 123:
```suggestion
```

liamhuber: You need this or the IDE complains later about defining an attribute outside init

liamhuber: No, I remember now. `IsNodal` doesn't use the property/setter obfuscation -- it just directly applies the init kwarg `self.parent = parent`. To conform to the spec that workflows shouldn't have parents, I wrap this by having an underlying `self._parent` attribute whose setter raises an error if you set it with anything other than `None`. Thus, we need to declare the private variable _before_ calling `super.__init__` since the super call will exploit the setter.

Since `parent` isn't exposed in `Workflow.__init__` and defaults to `None` in `IsNodal.__init__`, this _should_ all be redundant. But it keeps it all working together while ensuring that someone gets an error if they ever try to set `wf.parent = foo`.

In the future, if we want to change the spec so workflows _can_ have parents, all we need to do is get rid of the private var/setter/getter protection here (and maybe expose parent as a kwarg) and we could directly lean on the how super class handles the variable. I.e. it's really easy to unwind these protections if we change our mind about what the behaviour ought to be. There's a comment to this effect in the setter, but I'll expand it a bit

samwaseda: But `self._parent` is not used anywhere else other than this line right?

samwaseda:
uhm.... I didn't approve it yet...

pyiron_contrib/workflow/workflow.py
liamhuber: > But `self._parent` is not used anywhere else other than this line right?

Ah, after committing your other suggestion its not 🤦‍♂️ 

I added your suggestion over in #729

samwaseda: yeah the two suggestions were meant to be connected. I guess there's no way on GitHub to make one suggestion with two separate lines

samwaseda:
but ok now since `self._parent` got clarified I'm fine.

liamhuber:
> uhm.... I didn't approve it yet...

No worries, I'm happy to open a new PR with further changes you want. In the cases where you request explicit and significant changes that I disagree with, I'll hold off until we reach consensus. But if there's just a few comments and they're either small or I incorporate your feedback directly, then I'll probably be a bit pragmatic/hasty and merge them.

I often have hours every day to work on this, but that doesn't start until ~18:00 your time. It's not reasonable to expect you to keep reviewing and providing feedback consistently that late in the evening and I often want to keep building on stuff without worrying about huge merge headaches later, so I'd rather push it through and then make new PRs to fix anything you hate afterwards.

> but ok now since `self._parent` got clarified I'm fine.

Super! Yeah, this is what I'm talking about -- if the potential disagreement looks small and simple enough I'll probably just go for it and fix things later.

PR #748, opened 2023-06-28
Workflow: Composition to inheritance

liamhuber:
Transitions the `HasNodes` mix-in class  for `Workflow` (and later macros) to a `Composite(IsNodal)` parent class. This resolves some tension where classes demanding a `HasNodes` object actually sometimes expected said object to inherit from `IsNodal`. Also allows `Workflow` and macros to share methods (mainly `run`). Brings the code closer into line with the tentative spec in #743.


liamhuber:
The changes in the target branch #729 are no longer super meaningful by themselves, since this merges into that and all the stuff stacked on top of this got reviewed and/or approved. I'll merge this now, then we can have a meaningful conversation about the aggregate differences with `main`.

PR #749, opened 2023-06-28
Composition to inheritance

liamhuber:
Changes the `Node` type hints to `IsNodal` where appropriate. Had to touch up one small edge case in the `Workflow` tests as a result.


liamhuber:
Was trying to merge the type checking updates _into_ `composition_to_inheritance`. I have no idea how this got opened instead.

PR #751, opened 2023-06-28
Rename classes

liamhuber:
Rename `Node` to `Function` and then `IsNodal` to `Node`, including downstream changes like changing the decorator from `@node` to `@function_node`, the children `FastNode` to just `Fast`, etc.

In the process of updating the example notebook, I also made two small fixes to problems that had slipped through the cracks because they didn't raise any errors:
- Changing the `workflow=wf` kwarg to `parent=wf` in one of the instantiations
- Changing some of the update/run stuff in cell 17 to reflect what happens by default.

Not included in this PR: Moving all the docstrings around so that they better conform to reality and the spec in #743 -- right now `Function(Node)` has a bunch of pretty general specifications that should probably be moved over to `Node`, and the idea of macro nodes is not yet mentioned.


samwaseda:
Is there a possibility to run tests here? It looks fine but without tests it's difficult to say...

liamhuber:
> Is there a possibility to run tests here? It looks fine but without tests it's difficult to say...

We could rebase this onto main briefly, which will trigger tests on this (including everything this is stacked on top of), or we can change the centralized CI as you and I propose to do/discuss in the Monday meeting.

For each of this stack of PRs I made sure the workflow tests are passing on my local machine; it will still be necessary to do the full CI on other architectures, but I guess it's already quite promising.

liamhuber:
>  It looks fine but without tests it's difficult to say...

Since the tests are running on my machine and we're not merging into `main` here, I'm going to lean into the `it looks fine` part of that comment and boldly merge.

PR #753, opened 2023-06-28
Update docs

liamhuber:
It's not exhaustive or finalized, but in general the universal stuff now sits on `Node` and the stuff specific to wrapping functions sits on `Function(Node)`.


samwaseda (approved):
I didn't take a proper look, but I guess it's ok

PR #754, opened 2023-06-29
Data channel value starts as not data

### LINKED ISSUE #752: Add a `NotData` class to initialize data channel values
As discussed in #746 we should create a new `NotData` class, initializing `DataChannel.value = NotData()`, and updating the `DataChannel.ready` property to always return `False` when `isinstance(value, NotData)`. Then until the first time the channel is updated -- either by the `Node` providing a specified default, or by a user or graph connection passing new data -- the channel and thus node will not be ready.

This might free us up to make things like `run_on_updates` and `update_on_instantiation` default to `True` instead of `False`, because this new readiness check will prevent super simple nodes from running before they're ready and crashing!

I'd like to get #729 and everything stacked on top of it merged in first, since that's (more or less...) refactoring, and this is really a new feature.

liamhuber:
Introduces a new `NotData` class with which to initialize `DataChannel.default` and modify the channel readiness check so that it fails if the value is not data.

Closes #752.

Opens the door for having `run_on_updates` and `update_on_instantiation` both default to `True`, since now ill-defined functions won't automatically try to run and fail!


pyiron_contrib/workflow/channels.py
```diff
@@ -199,9 +212,13 @@ def ready(self) -> bool:
...
         """
         if self.type_hint is not None:
-            return valid_value(self.value, self.type_hint)
+            return self.value_is_data and valid_value(self.value, self.type_hint)
         else:
-            return True
+            return self.value_is_data
+
+    @property
+    def value_is_data(self):
```
samwaseda, line 220: Does it make maybe sense to make it private?

liamhuber: Yeah, just to avoid cluttering the tab completion list, sure.

I'm going to go have some breakfast, then I'll get this change in.

PR #755, opened 2023-06-29
Fast by default

liamhuber:
Now that we have `NotData` in place (#754), we can safely make the standard behaviour for `Function` nodes to be "fast" (running when input is updated). `SingleValue` is reparented right onto `Function`, and `Fast` is replaced by `Slow` (which is just a `Function` with different defaults for the run flags). This makes the standard `Function` behaviour more intuitive and streamlines some of the pedagogical stuff.


liamhuber:
@samwaseda, your reviews and PRs have made the `workflow` module _strictly and substantially better_, but there are many weeks where this is the main thing I've got to work on while I imagine you have a number of things on your plate -- so I want to be super conscious of not monopolizing your time, and I don't always want to wait for you to get to the reviews before continuing my work! I think stacking PRs is a good first-order solution to these problems, but I'm very open to having a discussion about our process in case you are tired of all the pings or have ideas for how we can make things even more efficient 😄

samwaseda (approved):

PR #761, opened 2023-07-06
Order things for circular graphs

liamhuber:
Had to update the order some data flows out of nodes. It's pretty rough, but an integration test shows how it works. Found a race condition sometimes -- we can fix it with the tools we have available (run signals) but it's annoying at a minimum and bad at a maximum.


pyiron_contrib/workflow/node.py
```diff
@@ -204,15 +204,14 @@ def finish_run(self, run_output: tuple):
         execution is finished.
         """
```
samwaseda, line 205:
```suggestion
        """
        self.running = False
```

pyiron_contrib/workflow/node.py
```diff
@@ -204,15 +204,14 @@ def finish_run(self, run_output: tuple):
...
         try:
+            self.running = False
```
samwaseda, line 207:
```suggestion
```

pyiron_contrib/workflow/node.py
```diff
@@ -204,15 +204,14 @@ def finish_run(self, run_output: tuple):
...
             self.process_run_result(run_output)
+            self.signals.output.ran()
         except Exception as e:
             self.running = False
```
samwaseda, line 211:
```suggestion
```

samwaseda (changes requested):

samwaseda:
The changes are connected

liamhuber:
> The changes are connected

Agreed, that reads much nicer. All committed.

samwaseda (approved):

PR #765, opened 2023-07-10
Executor submodule

liamhuber:
Introduce a new executor submodule for our custom children of `concurrent.futures.Executor`. Once it's wrapped up in class form I'd like to stick @jan-janssen's hash-and-serialize executor here so I can start playing with it in the workflows. Jan, if I get to packaging it that way first (maybe tomorrow or Wednesday if life goes smoothly?) I'll open a draft PR right away so you can see what's going on. 

In the meantime, here is a little executor I found on stack overflow that exploits cloudpickle so that we can handle dynamically defined classes. I did a little MWE of how workflow node decorators are running and test against that.

### DIFF
--- a/pyiron_contrib/executors/executors.py
+++ b/pyiron_contrib/executors/executors.py
@@ -10,13 +10,17 @@ def _apply_cloudpickle(fn, /, *args, **kwargs):
 
 class CloudpickleProcessPoolExecutor(ProcessPoolExecutor):
     """
-    This wrapper class replaces the `pickle` backend of the
-    `concurrent.futures.ProcessPoolExecutor` with a backend from `cloudpickle`.
-    In this way, even objects with no canonical import (e.g. those created dynamically
-    from a decorator) can be submitted to the executor.
+    This executor behaves like `concurrent.futures.ProcessPoolExecutor`, except that
+    non-pickleable callables may also be submit (e.g. dynamically defined functions).
+
+    This is accomplished by replacing the `pickle` backend of the
+    `concurrent.futures.ProcessPoolExecutor` with a backend from `cloudpickle` when
+    serializing the callable.
 
     This solution comes from u/mostsquares @ stackoverflow:
     https://stackoverflow.com/questions/62830970/submit-dynamically-loaded-functions-to-the-processpoolexecutor
+
+    Note: Arguments and return values must still be regularly pickleable.
     """
 
     def submit(self, fn, /, *args, **kwargs):
--- a/pyiron_contrib/executors/executors.py
+++ b/pyiron_contrib/executors/executors.py
@@ -1,0 +1,25 @@
+from concurrent.futures import ProcessPoolExecutor
+
+import cloudpickle
+
+
+def _apply_cloudpickle(fn, /, *args, **kwargs):
+    fn = cloudpickle.loads(fn)
+    return fn(*args, **kwargs)
+
+
+class CloudpickleProcessPoolExecutor(ProcessPoolExecutor):
+    """
+    This wrapper class replaces the `pickle` backend of the
+    `concurrent.futures.ProcessPoolExecutor` with a backend from `cloudpickle`.
+    In this way, even objects with no canonical import (e.g. those created dynamically
+    from a decorator) can be submitted to the executor.
+
+    This solution comes from u/mostsquares @ stackoverflow:
+    https://stackoverflow.com/questions/62830970/submit-dynamically-loaded-functions-to-the-processpoolexecutor
+    """
+
+    def submit(self, fn, /, *args, **kwargs):
+        return super().submit(
+            _apply_cloudpickle, cloudpickle.dumps(fn), *args, **kwargs
+        )
--- a/pyiron_contrib/executors/README.md
+++ b/pyiron_contrib/executors/README.md
@@ -1,0 +1,3 @@
+# Executors
+
+This sub-module holds custom children of `concurrent.futures.Executor` for use in other parts of pyiron (e.g. `pyiron_contrib.workflow`).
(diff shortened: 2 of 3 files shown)


pyiron_contrib/executors/executors.py
```diff
@@ -0,0 +1,25 @@
+from concurrent.futures import ProcessPoolExecutor
+
+import cloudpickle
+
+
+def _apply_cloudpickle(fn, /, *args, **kwargs):
+    fn = cloudpickle.loads(fn)
+    return fn(*args, **kwargs)
```
jan-janssen, line 8: I guess you also want to `cloudpickle` the output on the way back.

jan-janssen:
Just for future reference https://github.com/pyiron-dev/pysqa-executor

pyiron_contrib/executors/executors.py
liamhuber: I'm not sure what you mean. Can you suggest a modification to the test that makes it fail without this?

jan-janssen, line 8:
```suggestion
    return cloudpickle.dumps(fn(*args, **kwargs))
```

jan-janssen: I added a suggestion below. The other part I am surprised about is that you do not pickle the arguments, but only the function. Should not we pickle everything together?

liamhuber: I guess this is only necessary when the args or return values are themselves things that cannot be normally pickled (eg instances of classes created by a decorator). Tomorrow I'll come back and make the test case much crueler. Hopefully then we'll see the tests fail without your suggestion and similar treatment for the args

pmrv (approved):
Looks ok to me to get the ball rolling. Can we think about making this dynamic somehow?  Like I give you any executor and you give me back one that cloudpickles?  OTOH, we'll on the order of less than a dozen "official" executors, so wrapping them once is also not too much hassle. 

Since we have this now, I'll go ahead and rename tinybase's executors, since their are slightly different and I didn't like that name clash to begin with.

liamhuber:
I'm working on this currently. I agree that for non-pickleable callables that take pickleable arguments and have a pickleable return value, it's fine. Jan's suggestion fixes the case of non-pickleable return values _but_ then the future results need to be unpickled post-facto, i.e. `type(fs.results())` is `<class 'bytes'>` while `coudpickle.loads(fs.results())` is the thing we actually want.

I see a potential solution where we really replace the underlying `Future` class with one whose `.results()` super's the parent class but then un-cloud-pickles any bytes found. But this is getting quite extreme. I'm going to go have lunch and think about it a bit.

liamhuber:
So I'd like to merge this one as-is. I added a bit of stuff so that both the docstrings and tests are more explicit that this handles unpickleable _callables_ but that arguments and return values must still be pickleable. This is already totally sufficient for #760, where our callables are methods of dynamically created nodes, but the input and output is (so far) always just regular data.

Jan's suggestion above _does_ extend the class to handle unpickleable return values, and one can imagine a similar attack on arguments. The problem is that it also messes with the interface, since `fs.result()` is now a bytecode object in need of `cloudpickle.loads` instead of just the actual result we want.

I have a solution in mind (sketched below), but I would much rather get the current code -- which is working fine and tested -- merged, and either replace it or add a new more powerful executor in a future PR. This will keep the PRs smaller, give Jan's [serializing and hashing executor](https://github.com/pyiron-dev/pysqa-executor) a logical place to live on contrib, and remove barriers to working on #760 in case my sketch of an idea doesn't pan out.

The idea:

We can get things working ok, but the future result is ugly, so we'll need a new class to handle this, like

```python
class CloudLoadsFuture(Future):
    def result(self):
        result = super().result()
        if isinstance(result, bytes):
            result = cloudpickle.loads(result)
        return result
```

Then we apply a wrapper and subclass the subprocess executor to make sure everything is getting serialized _before_ it gets seen by the `multiprocessing` library:

```python
class CloudPickledCallable:
    def __init__(self, fnc: callable):
        self.fnc_serial = cloudpickle.dumps(fnc)

    def __call__(self, /, dumped_args, dumped_kwargs):
        fnc = cloudpickle.loads(self.fnc_serial)
        args = cloudpickle.loads(dumped_args)
        kwargs = cloudpickle.loads(dumped_kwargs)
        return cloudpickle.dumps(fnc(*args, **kwargs))

    @classmethod
    def dumps(cls, stuff):
        return cloudpickle.dumps(stuff)


class CloudpickleProcessPoolExecutor(ProcessPoolExecutor):
    def submit(self, fn, /, *args, **kwargs):
        return self._submit(
            CloudPickledCallable(fn), 
            CloudPickledCallable.dumps(args), 
            CloudPickledCallable.dumps(kwargs)
        )

    def _submit(self, fn, /, *args, **kwargs):
        # This is just exactly ProcessPoolExecutor.submit, but we hack in
        # f = CloudLoadsFuture()
        # instead of 
        # f = _base.Future()
        # for the future instantiation
        raise NotImplementedError
```

Of course this may still not do what I want, is probably not the most elegant solution, and I want to take another peek around to see if there is an existing tool that does this (there's at least a `multiprocessing` replacement [`multiprocess`](https://pypi.org/project/multiprocess/) that replaces pickle with dill...), but these are all reasons I'd rather defer this more complex (and as yet unnecessary) task to another PR.

liamhuber:
UPDATE: Actually that plan of attack worked basically flawlessly (at least as far as my current tests can see), and is drafted over in #767

jan-janssen:
> @jan-janssen, are you amenable to such a split/delay?

As it is `pyiron_contrib` I am fine with merging this as is. @pmrv and I had a similar discussion in https://github.com/pyiron/pyiron_contrib/issues/766 and we both agreed that the `Executors` still need some extensions before we can use them as the core for the next generation of `pyiron_base`. Nevertheless this should not block the development here in `pyiron_contrib` and it helps us to clarify what we still need to work on.

Issue #756, opened 2023-06-30
Suggestions and issues for the workflow class


JNmpi:
@liamhuber, thanks for all the new features and enhancements for the workflow class. In the following, I list some issues and suggestions regarding the workflow class. For the following jupyter notebook I used the main branch. Thus, some of the issues may have been fixed in the latest branch.

```python
from pyiron_contrib.workflow.node import Node, node, fast_node
from pyiron_contrib.workflow.workflow import Workflow
```

#### Using the decorator for a single output variable fails

```python
@node('y')
def identity(x: int=0) -> int:
    return x

node_identity = identity(0)
# node_identity.run()  # uncomment this line to see the error message
```

#### The following construction without a decorator works as expected

```python
def identity(x: int=0) -> int:
    return x

node_identity = Node(identity, 'y')

node_identity.inputs.x = 10
node_identity.run()
node_identity.outputs.y.value

```

    10

#### Some suggestions for making working with the nodes more convenient

Return the available input channels (same for outputs):

```python
node_identity.inputs
```

    <pyiron_contrib.workflow.io.Inputs at 0x16b37de10>

Return the output (e.g. node.outputs.to_value_dict()), similar like dask.compute() when performing run:

```python
node_identity.run()
```

#### Connecting two nodes

```python
@node("p1", "m1")
def my_mwe_node(
    x: int | float, y: int | float = 1) -> tuple[int | float, int | float]:
    return x+1, y-1

node_instance = my_mwe_node(x=0)
node_instance.inputs.y = node_identity.outputs.y

node_identity.outputs.y.connected
```

    True

The following code should 'see' that the input of the first node has been updated (i.e., should print 5 rather than 10):

```python
node_identity.inputs.x = 5
#node_identity.run()
print (node_instance.inputs.y)
print (node_identity.outputs.y)
node_instance.run()
```

    10
    10

To get the expected behavior I have to manually run the first node (which is not what I would like to have for a delayed operation):

```python
node_identity.inputs.x = 5
node_identity.run()
print (node_instance.inputs.y)
print (node_identity.outputs.y)
node_instance.run()
```

    5
    5

```python

```

liamhuber:
Thanks for the testing and feedback @JNmpi!
 fixed in the latest branch.

Here's a summary of the TODO's I see until this issue can be closed:
- [ ] Raise a cleaner error when nodes created by decoration erroneously receive `*output_labels` at instantiation
[...]

PR #767, opened 2023-07-11
More cloudpickle executor

liamhuber:
A quick draft of the idea discussed in #765 to use cloudpickle for _everything_ being passed to the `ProcessPoolExecutor` (callable, args, and return value). I've got to stop for the day already, but the first attack got working so I wanted to push it.

Todo:
- [ ] wait for #756 to be merged
- [ ] Rename the public facing classes
- [ ] Expand tests
- [ ] Add docstrings

Should this just supplant the simpler executor that only cloudpickles the callable? I lean towards "no", since I suspect this is slower and may be overkill a lot of the time, but I'm open to dissent and want to do some timing tests before a final decision.

### DIFF
--- a/pyiron_contrib/executors/README.md
+++ b/pyiron_contrib/executors/README.md
@@ -1,0 +1,3 @@
+# Executors
+
+This sub-module holds custom children of `concurrent.futures.Executor` for use in other parts of pyiron (e.g. `pyiron_contrib.workflow`).
--- a/pyiron_contrib/executors/executors.py
+++ b/pyiron_contrib/executors/executors.py
@@ -1,4 +1,5 @@
-from concurrent.futures import ProcessPoolExecutor
+from concurrent.futures import Future, ProcessPoolExecutor
+from concurrent.futures.process import _global_shutdown, _WorkItem, BrokenProcessPool
 
 import cloudpickle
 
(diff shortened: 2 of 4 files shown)


liamhuber:
I did some speed testing and don't see a significant slowdown from using the more complete executor that cloudpickles _everything_ instead of just the called function. For simplicity, I'll just get rid of the more restrictive class.

Test:
```python
from functools import partialmethod
from sys import getsizeof
import time

from pyiron_contrib.executors.executors import CloudProcessPoolExecutor, CloudpickleProcessPoolExecutor

class Foo:
    """
    A base class to be dynamically modified for testing CloudpickleProcessPoolExecutor.
    """
    def __init__(self, fnc: callable):
        self.fnc = fnc
        self.result = None

    @property
    def run(self):
        return self.fnc

    def process_result(self, future):
        self.result = future.result()


def dynamic_foo():
    """
    A decorator for dynamically modifying the Foo class to test
    CloudpickleProcessPoolExecutor.

    Overrides the `fnc` input of `Foo` with the decorated function.
    """
    def as_dynamic_foo(fnc: callable):
        return type(
            "DynamicFoo",
            (Foo,),  # Define parentage
            {
                "__init__": partialmethod(
                    Foo.__init__,
                    fnc
                )
            },
        )

    return as_dynamic_foo

@dynamic_foo()
def fnc(x):
    return x
f = fnc()

arg = {n: "foo"*n for n in range(10000)}  # or arg = 1
print(getsizeof(arg)/1000000, "MB")
>>> 0.294992 MB
```

```python
%%timeit
ex = CloudpickleProcessPoolExecutor()  # Just the callable
fs = ex.submit(f.run, arg)
fs.result()
>>> 13.6 s ± 168 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

```python
%%timeit
ex = CloudProcessPoolExecutor()  # Callable, args, and return value
fs = ex.submit(f.run, arg)
fs.result()
>>> 14 s ± 134 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

So _just_ outside agreeing within error. When the argument is small (`1`), they agree within error and sometimes the more extensive cloudpickler comes out ahead.

liamhuber:
linux 3.8 error reads like this in a few places:

```
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/pyiron_contrib/pyiron_contrib/tests/unit/executor/test_cloudprocesspoolexecutor.py", line 99, in test_unpickleable_return
    fs = executor.submit(dynamic_dynamic.run)
  File "/home/runner/work/pyiron_contrib/pyiron_contrib/pyiron_contrib/executors/executors.py", line 58, in submit
    return self._submit(
  File "/home/runner/work/pyiron_contrib/pyiron_contrib/pyiron_contrib/executors/executors.py", line 81, in _submit
    self._executor_manager_thread_wakeup.wakeup()
AttributeError: 'CloudProcessPoolExecutor' object has no attribute '_executor_manager_thread_wakeup'

----------------------------------------------------------------------
Ran 93 tests in 11.308s

FAILED (errors=5, skipped=40)
```

This is not the end of the world. I'm directly breaking into `concurrent.futures.ProcessPoolExecutor.submit` and replacing `f = _base.Future()` with our future and just copy-pasting the rest of the method. They've simply got different attributes in different versions. It would be nice to find a more elegant solution than copying the rest of the method directly, but it is at least simple to fix this by having a method for each version.

Issue #717, opened 2023-06-16
Workflow: Automatic parsing of outputs


samwaseda:
I saw [this PR](https://github.com/pyiron/pyiron_contrib/pull/715) and I have to admit that it is not only cumbersome to have to write all the output names, but also it is potentially super dangerous if the user has to check the consistency of outputs. The suggestion below is not immediately applicable in that PR, but I think it would allow for a better systematic recognition of output variables:

```python
import ast
import inspect
import re


def remove_spaces_until_character(string):
    pattern = r'\s+(?=\s)'
    modified_string = re.sub(pattern, '', string)
    return modified_string


class ParseOutput:
    def __init__(self, function):
        self._func = function
        self._source = None
        
    @property
    def func(self):
        return self._func
    
    @property
    def node_return(self):
        tree = ast.parse(inspect.getsource(self.func))
        for node in ast.walk(tree):
            if isinstance(node, ast.Return):
                return node

    @property
    def source(self):
        if self._source is None:
            self._source = [l.rsplit("\n", 1)[0] for l in inspect.getsourcelines(self.func)[0]]
        return self._source
    
    def get_string(self, node):
        string = ""
        for ll in range(node.lineno - 1, node.end_lineno):
            if ll == node.lineno - 1 == node.end_lineno - 1:
                string += remove_spaces_until_character(self.source[ll][node.col_offset:node.end_col_offset])
            elif ll == node.lineno - 1:
                string += remove_spaces_until_character(self.source[ll][node.col_offset:])
            elif ll == node.end_lineno - 1:
                string += remove_spaces_until_character(self.source[ll][:node.end_col_offset])
            else:
                string += remove_spaces_until_character(self.source[ll])
        return string
    
    @property
    def output(self):
        if self.node_return is None:
            return
        if isinstance(self.node_return.value, ast.Tuple):
            return [self.get_string(s) for s in self.node_return.value.dims]
        return [self.get_string(self.node_return.value)]
```

What it does:

```python
def identity(x):
    return x

print(ParseOutput(identity).output)

def add(x, y):
    return x + y

print(ParseOutput(add).output)

def add_and_subtract(x, y):
    return x + y, x - y

print(ParseOutput(add_and_subtract).output)

def md(job):
    temperature = job.output.temperature
    energy = job.output.energy
    return temperature, energy

print(ParseOutput(md).output)

def function_return(i, j):
    return (
        np.arange(
            i, dtype=int
        ),
        np.shape(i, j)
[...]

PR #769, opened 2023-07-12
Node input as args

liamhuber:
So far, this modifies `Function.__call__` such that it updates the input channels using `**kwargs` _and/or_ `*args`. This is the first step towards being able to _instantiate_ nodes (at least those created by a decorator) with their initial input values set as args, as requested by @JNmpi in #756.


liamhuber:
I let this idea percolate overnight, and I think the issue is that in comparison to Joerg's demo in #757, our `Function(Node)` instances have their own important `Node` kwargs _in addition_ to the arguments of the underlying function being wrapped. So in both instantiation _and_ at call time we need to parse out which kwargs are being used to update node properties, and which are being used to update input channels.

Of course, no such division of `*args` is sensibly possible -- if the wrapped function accepts strings as output, how could we ever know if a particular arg is meant as an output channel label or input data? For this reason, before working further on this PR, I'd like to go back and implement @samwaseda's idea to automatically parse output labels in #717. Of course we can still allow custom output labels, but they'd be from a kwargs `output_labels: Optional[tuple | list] = None` instead of the args `*output_labels`. Then we can rebase this PR on top of that and work in a world where `*args` are _always_ interpreted as input data... at least for `Function(Node)`, since for `Composite(Node)` there is no explicit order to the input so we'll still need to do input data updates with kwargs.

JNmpi:
Thanks, @liamhuber for your thoughts. I am not sure that I understand the issue of parameters coming from the function we want to wrap and those related to the node. The strings needed for the output label are arguments in the decorator, whereas all function arguments are contained in the function definition. Thus, when converting the function via a decorator the two sets of parameters should be straightforward to distinguish. Or is there something I am missing? It may be helpful to give a code example.

liamhuber:
Hi @JNmpi, the problem was that the strings for the output labels were provided to the decorator and/or the underlying node class at instantiation. Since there's no robust way to distinguish whether an `*arg` is intended as a label (arg of the class) or function input (arg of the function the class is wrapping), so we were really stuck. Now that output labels are automatically parsed or provided as a kwarg, `*args` are _always_ function input and we're a-ok.

The other comment was just that we want to be able to pass, eg, `run_on_updates=True` at instantiation (and on call?) and recognize that this is a node class kwarg and not wrapped function input. This is not a problem at all and we already have all the tools to handle it; I should be able to get it put together on Monday.

liamhuber:
@samwaseda, @JNmpi, there is of course more that can and will be added, but #770 and this PR together make something I'm pretty happy with. We now have a lot more freedom in terms of (not) labelling output channels, and updating input at init and on call with both args and kwargs. Example:

```python

wf = Workflow("simple")

@Workflow.wrap_as.single_value_node()
def add_one(x):
    y = x + 1
    return y  # Output labels parsed automatically

@Workflow.wrap_as.function_node(output_labels="sum")  # Output labels optionally overridden
def adder(x: int | float = 1, y: int | float = 1) -> int | float:
    return x + y

# Function node input data as args, kwargs, or both!
wf.a = add_one(0)   # arg
wf.b = add_one(x=0)   # kwarg
wf.sum = add_node(wf.a, y=wf.b)   # both
# Remember, with single value nodes we can pass the whole node instead of an output channel!

# Workflow automatically generate {node.label}_{channel.label} IO
print(wf.outputs.sum_sum.value)
>>> 2

# Can call a workflow, but since input has no order _only_ kwargs work
wf(a_x=2, b_x=3)
print(wf.outputs.sum_sum.value)
>>> 7
```

Docs, tests, and example notebook are all updated accordingly.

samwaseda:
Sorry I cannot review until Wednesday evening!!

liamhuber:
> Sorry I cannot review until Wednesday evening!!

No worries! I've got extra childcare stuff today anyways; if I do manage to get any more done I'll just keep stacking it on top.

PR #770, opened 2023-07-13
Parse output labels automatically 

### LINKED ISSUE #717: Workflow: Automatic parsing of outputs

liamhuber:
Closes #717 

First, this makes it easier to define new nodes, since there will just be less to write. Second, this will let us free up `*args` in `Function(Node)` to be used for providing input data (as discussed in #769). We'll still allow explicitly providing different output labels, but this becomes an optional kwarg instead of mandatory args. 

Best-practice becomes to give good variable names to all the return values in your wrapped function, but Sam's parser is super generic and you'll still be able to access weirdly-named nodes with `["node name"]` even if they're not dot-accessible. (At least that's the plan!)

- [x] Implement and test the function return value parser from @samwaseda in #717 
- [x] Use the parser for `Function(Node)` and its children
- [x] Update all the examples and docs


liamhuber:
Pip-check looks unrelated:

```
cfn-lint 0.78.1 has requirement aws-sam-translator>=1.70.0, but you have aws-sam-translator 1.68.0.
```

liamhuber:
@samwaseda, should be good now. `output_labels` can still be passed as a kwarg, which allows you totally override the label scraping. This is helpful when returned values have crappy names, and necessary when (a) you _want_ a single channel carrying a tuple instead of decomposing the output or (b) the wrapped function has multiple `return` expressions (we might _sometimes_ be able to handle this, e.g. extra return calls from functions defined inside the wrapped function, but it's an edge-enough case that I don't mind forcing the user to provide one more kwarg when it comes up.)

The pip failure seems unrelated to me, but let's figure out where it's from next week and fix it before merging.

Excellent idea with this automatic parsing btw, I'm really happy with the improvement in usability.

PR #777, opened 2023-07-18
Clean executor fails

liamhuber:
Running on an executor is giving me `TypeError: cannot pickle '_thread.lock' object`, i.e. for `Function` nodes that use `self` and for `Composite` nodes we can't yet use an executor. Until this is fixed, it should fail cleanly and informatively.


liamhuber:
@samwaseda you're welcome to review this, but IMO it's low priority as I'm just making sure we have cleaner errors in an edge case.

There's a chance that Jan's `pympipool.Executor` can handle this situation already, so I'll try that before really digging in to solve this from our end. Fingers crossed it will be easy!

PR #782, opened 2023-07-19
Run the workhorse CI on _all_ pulls

liamhuber:
Not just those merging to main.

We talked about this at the meeting and I don't remember any real pushback, so let's try it out in contrib.

Normally we rename files in totally separate PRs, but this is so trivial I thought I could get away with both at once. The actual material changes are easily visible by looking at the first commit.

If we like this we can go upstream to pyiron/actions and similarly rename things/change the commented-out advice at the head of reusable workflows.


niklassiemer (approved):
Lgtm

Issue #764, opened 2023-07-10
Node granularity - When do we consider a node to be macro?


mbruns91:
After our discussion on Monday, I was thinking more about the concept of node granularity and I actually couldn't come up with a 'natural' way to define something I'd call an 'atomic node'. An atomic node is for me a node which cannot be decomposed any further into sub-nodes. Taking the example from @JNmpi of a numpy linspace node (as far as I recall correctly),
```python
@np_linspace
def linspace() -> np.array[int]:
    return np.linspace()
```
I'd suppose that's as atomic as we should reasonably go. One could, however, decompose this further by looking into numpy's internals, see how linspace is exactly executed, create a node for each of these steps and voilà, the numpy linspace node could also be constructed as a composite node (which is imho a better term than "macro node"). Thus we have to come up with definitions for the levels of **node-granularity** and in this context **node types**.  
I'd start with the terms I already used:
- `atomic_node`: a (in our framework) node, which cannot be further decomposed
- `composite_node`: a node which is composed of atomic nodes
Above example also illustrates that this will be important for node-developers (and maybe node-store maintainers?) as it can easily happen that nodes that were considered to be atomic during development turn out to be composite. Also this can change over time, when e.g. more atomic nodes are developed, making older, more complex atomic nodes actually composite. Perhaps we even need more than two node types or more types of macro nodes (of which `composite_node` would then be a subclass)? What are you guy thinking about this?

liamhuber:
This is a good question! Right now we err on the side of extreme pragmatism: a node is a macro (i.e. `Composite`) iff it has its own internal graph structure, i.e. it owns a collections of other `Node` objects and its execution means executing these. Meanwhile an atomic, or `Function(Node)` is one that has no such internal structure, but executes arbitrary python code.

So we have no concept of an `atomic_node` that is "unable" to be further decomposed, not is the `composite_node` restricted to only be "composed of atomic nodes". Rather, by fiat a `Function` node executes python code, and a `Composite` node executes other nodes.

So the question of "when do we consider a node to be macro" becomes extremely simple and has a technical answer, and the responsibility "_should_ I implement this as a functional node or a macro node?" gets devolved to individual node developers. This also allows, from a node _user_ perspective, for a particular node to easily switch from functional to composite under the hood without changing its public interface. I'm certainly open to criticism of this paradigm or alternatives, but am so far very happy with it.

[...]

PR #786, opened 2023-07-19
Drawing the workflow with graphviz

liamhuber:
- [x] Draw node IO and distinguish data and signals channels
- Handle composite nodes
  - [x] Override `Composite.draw` to take a `granularity` keyword; when `granularity =0`, just draw IO like a regular node
  - [x] When `granularity ` > 0, draw owned nodes and their connections, with connections to map the composite IO channels to internal node channels
  - [x] Try/Except recursively calling children with granularity to get expandable node maps [per @mbruns91's suggestion](https://github.com/pyiron/pyiron_contrib/issues/764#issuecomment-1630999123)
- [x] Attempt to get things to line up good -- e.g. input and output channels top-to-bottom and everything else left-to-right (this might not be possible with graphviz, as it seems to be ignoring my current attempts to use `rankdir`, which AFAIK is the tool for this job)
- UX Polish
  - [x] Colour and/or shape different elements to reflect what they are (e.g. Composite/Function/SingleValue/Slow, data and signals being differently coloured, etc.)
  - [x] allow a vertical or horizontal flag to reverse the rankdir difference in nodes and IO
  - ~For unconnected data channels, make a little box showing the current value feeding into/out of it~ This is actually a bit tricky for all data that doesn't have a nice (and short!!!!) representation, so let's just leave it for now.
- [x] Update docs
- [x] Update example notebook

# As of writing...

This graph:

```python

wf = Workflow("with_prebuilt")

wf.structure = wf.add.atomistics.BulkStructure(repeat=3, cubic=True, element="Al")
wf.engine = wf.add.atomistics.Lammps(structure=wf.structure)
wf.calc = wf.add.atomistics.CalcMd(job=wf.engine)
wf.calc.signals.input.run = wf.engine.signals.output.ran
```

Gives stuff that looks like

```python
wf.structure.draw()
```

structure

and

```python
wf.draw()  # Default granularity = 1
```

wf


samwaseda:
Sorry today I'm super busy but I'll try to find time tomorrow!!

PR #793, opened 2023-07-24
[functionality] QOL utils (parallelised vasp import, compression, decompression etc.)

ligerzero-ai:
Quality of life utilities, including but not limited to:

- Find folders with specified files
- Extract tarballs
- Compress tarballs
- Import vasp calculations into a df _in parallel_ via a single call with function

e.g.
```python
from pyiron_contrib.utils.vasp import DatabaseGenerator
import argparse

def main():
    parser = argparse.ArgumentParser(description='Find and compress directories based on specified criteria.')
    parser.add_argument('directory', metavar='DIR', type=str, help='the directory to operate on')
    args = parser.parse_args()
    
    datagen = DatabaseGenerator(args.directory)
    df = datagen.build_database(max_dir_count = 2000)

if __name__ == '__main__':
    main()
```
```shell
python3 /home/562/hlm562/python_scripts/build_vasp_database.py $PWD >> py.output
```
- Chargemol utils for plotting bonds/bond analysis

Some notes:

- [ ] Need to switch to `pyfileindex` for file search via os.walk
- [ ] Need to switch to `pympipool` when I get around to it.
- [ ] Adjust import statements to include `pyiron_contrib`
- [ ] Add option to specify workers in `parallelise`


ligerzero-ai:
@jan-janssen 

Notes:
There is no infrastructure for adding functions similar as to a pyiron table - this is purely for my own purposes which is to import vasp calculations in parallel. 

There will need to be some conceptualisation of what the table functionality looks like as a standalone module. But a lot of the utilities can help with that, I imagine.

PR #800, opened 2023-08-02
Downgrade pyparsing

srmnitc:
To fix https://github.com/conda-forge/pyiron_contrib-feedstock/pull/32

### DIFF
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -10,7 +10,7 @@ dependencies:
 - numpy =1.24.3
 - pyiron_base =0.6.3
 - pyiron_atomistics =0.3.0
-- pyparsing =3.0.9
+- pyparsing <3.1
 - scipy =1.10.1
 - seaborn =0.12.2
 - scikit-image =0.21.0
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -10,7 +10,7 @@ dependencies:
 - numpy =1.24.3
 - pyiron_base =0.6.3
 - pyiron_atomistics =0.3.0
-- pyparsing <3.1
+- pyparsing =3.0.9
 - scipy =1.10.1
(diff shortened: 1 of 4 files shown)


jan-janssen:
Should not we just update `matplotlib` to again be compatible to `pyiron_atomistics`? From my perspective we can now disable python 3.8 on `pyiron_contrib` and then we should be able to update the open dependency bot pull requests.

srmnitc:
https://github.com/conda-forge/matplotlib-feedstock/blob/main/recipe/meta.yaml seems to still have pyparsing constraint.

.binder/environment.yml
```diff
@@ -10,7 +10,7 @@ dependencies:
 - numpy =1.24.3
 - pyiron_base =0.6.3
 - pyiron_atomistics =0.3.0
-- pyparsing =3.1.1
+- pyparsing <3.1
```
jan-janssen, line 13:
```suggestion
- pyparsing =3.0.9
```

setup.py
```diff
@@ -38,7 +38,7 @@
         'pyiron_base==0.6.3',
         'scipy==1.10.1',
         'seaborn==0.12.2',
-        'pyparsing==3.1.1'
+        'pyparsing<3.1'
```
jan-janssen, line 41:
```suggestion
        'pyparsing==3.0.9',
```

jan-janssen:
> https://github.com/conda-forge/matplotlib-feedstock/blob/main/recipe/meta.yaml seems to still have pyparsing constraint.

The issue is that these constraints were only introduced in `matplotlib =3.7.2` so by removing Python 3.8 support and updating `matplotlib` to `3.7.2` we should be able to spot the issue here in this repository before we make a new release.

jan-janssen (approved):
Looks good to me

PR #805, opened 2023-08-03
Better workflow representation

### LINKED ISSUE #756: Suggestions and issues for the workflow class

### LINKED ISSUE #804: JSON repr not always working
The `HasToDict` mix-in class used throughout the `workflow` module adds a custom `_repr_json_` method to get nice notebook representations for the various objects. Unfortunately, this isn't working for a couple of objects (originally discussed in PR #658).

M(ish)WE:

```python

wf = Workflow("with_prebuilt")

wf.structure = wf.add.atomistics.BulkStructure(repeat=3, cubic=True, element="Al")
wf.engine = wf.add.atomistics.Lammps(structure=wf.structure)
wf.calc = wf.add.atomistics.CalcMd(job=wf.engine)
wf.calc.signals.input.run = wf.engine.signals.output.ran
```

- `wf` is not good (`Workflow(Node)`)
- `wf.structure` is fine (`SingleValue(Function(Node))`)
- `wf.structure.inputs` is no good (`Inputs`)
- `wf.structure.inputs.cubic` is fine (`DataChannel`)

As a temporary fix in the linked PR, I added a `repr_json()` method that just explicitly wrapped the returned dictionary in `_repr_json_` in a `IPython.display.JSON` call before returning it, and this always works fine -- so it's just when the notebook tries to do it's magic that we run into trouble.

I think I see the issue already (it has to do with overloading `__getattr__`), but I want the solution process documented. I hope to raise and close this in a sitting.

liamhuber:
Some custom string representations for workflow objects, and squashing the bug that was stopping `_repr_json_` from triggering automatically on a couple of classes.

Closes #804.
Closes #756 (there is still useful conversation there to think about as we keep working on this code, but this was the last of the todo items for immediate resolution).

### DIFF
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -6,7 +6,6 @@ dependencies:
 - coveralls
 - coverage
 - codacy-coverage
-- ipython
 - matplotlib =3.7.2
 - numpy =1.24.3
 - pyiron_base =0.6.3
--- a/.ci_support/environment.yml
+++ b/.ci_support/environment.yml
@@ -6,7 +6,6 @@ dependencies:
 - coveralls
 - coverage
 - codacy-coverage
-- ipython
 - matplotlib =3.7.2
 - numpy =1.24.3
 - pyiron_base =0.6.3
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -8,7 +8,6 @@ dependencies:
 - coveralls
 - coverage
 - codacy-coverage
-- ipython
 - matplotlib =3.7.2
 - numpy =1.24.3
 - pyiron_base =0.6.3
--- a/pyiron_contrib/workflow/composite.py
+++ b/pyiron_contrib/workflow/composite.py
@@ -238,7 +238,14 @@ class Composite(Node, ABC):
             super().__setattr__(label, node)
 
     def __getattr__(self, key):
-        return self.nodes[key]
+        try:
+            return self.nodes[key]
+        except KeyError:
+            # Raise an attribute error from getattr to make sure hasattr works well!
+            raise AttributeError(
+                f"Could not find attribute {key} on {self.label} "
+                f"({self.__class__.__name__}) or in its nodes ({self.nodes.keys()})"
+            )
 
     def __getitem__(self, item):
         return self.__getattr__(item)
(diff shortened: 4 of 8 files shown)


liamhuber:
@samwaseda, you're welcome to take a peek at the string representations and error messages to let me know if you think there's better wording/presentation, but this is mostly just the error type bugfix, so I'm going to merge without review.

PR #832, opened 2023-08-30
Shell commands

liamhuber:
A first attack at (meta) node(s) for running shell commands via the workflows.

I'm relatively content with the `ShellCommand` node, although certainly open to criticism and extensions, e.g. thinking about how such a node is going to interact with executors.

As a first extension, I tried using `ShellCommand` as the core of a meta-node that runs executable python scripts with CLI args and kwargs.
I absolutely hate the implementation right now -- it's ugly as sin -- but the UI is not so bad.
@srmnitc, please take a peek at the user-facing side of things in the new integration test and let me know what you think! (Of course you're very welcome to give feedback on the implementation too, but that's still a hot mess and I don't demand that you subject yourself to that 🤣)

### DIFF
--- a/pyiron_contrib/workflow/meta.py
+++ b/pyiron_contrib/workflow/meta.py
@@ -372,7 +372,8 @@ def python_script(
 
         for process_output in macro.script.outputs.labels:
             macro.outputs_map[
-                f"{macro.script.label}__{process_output}"] = process_output
+                f"{macro.script.label}__{process_output}"
+            ] = process_output
 
     return type(
         script.title().replace("_", "").replace(" ", ""),  # fnc_name to CamelCase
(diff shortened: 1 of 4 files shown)


srmnitc:
Thanks, It looks good! I have not executed any code, but looks at it, the `python_script` node looks very good. I like how you can let the workflow know that a file created is an output. One minor thing that comes to my mind is to maybe rename `kwargs`. I have no issue with it, but I wonder if it could be well known outside the python community. 

I see there is a `shell_command` node. We could rename this as maybe as  `command_line_tool`, similar to CWL & co. Additionaly, a `shell_script` could be useful, as it sounds, can run a longer script; while `command_line_tool` runs a software directly.

PR #839, opened 2023-09-09
Return connection pairs on disconnection

liamhuber:
When breaking connections, return the broken connection as a tuple of channels (or a list of such tuples in the case of IO panels and nodes).

I think this doesn't hurt in terms of giving developers more access to what's going on under the hood, but in particular I have in mind to use it for workflows when they are automatically building signals connections to run workflows who's data flow conforms to a DAG. This change stands OK on its own though, so I thought I'd take the opportunity to make a reasonably-scoped PR while I can...


pyiron_contrib/workflow/channels.py
```diff
@@ -87,24 +87,31 @@ def connect(self, *others: Channel) -> None:
...
         Args:
             *others (Channel): The other channels to disconnect from.
+
+        Returns:
+            [list[tuple[Channel, Channel]]]: A list of the pairs of channels that no
+                longer participate in a connection.
         """
+        destroyed_connections = []
         for other in others:
             if other in self.connections:
```
samwaseda, line 104: Would it maybe deserve a warning if it's not found? This kind of stuff helps debugging

pyiron_contrib/workflow/io.py
```diff
@@ -226,9 +241,21 @@ def __init__(self):
...
+        Disconnect all connections in input and output signals.
+
+        Returns:
+            [list[tuple[Channel, Channel]]]: A list of the pairs of channels that no
+                longer participate in a connection.
+        """
+        destroyed_connections = []
+        destroyed_connections.extend(self.input.disconnect())
+        destroyed_connections.extend(self.output.disconnect())
+        return destroyed_connections
```
samwaseda, line 255:
```suggestion
        return self.input.disconnect() + self.output.disconnect()
```

pyiron_contrib/workflow/io.py
```diff
@@ -206,6 +215,12 @@ class InputSignals(SignalIO):
     def _channel_class(self) -> type(InputSignal):
         return InputSignal
 
+    def disconnect_run(self) -> list[tuple[Channel, Channel]]:
+        try:
+            return self.run.disconnect_all()
```
samwaseda, line 220: Where does this `run` come from?

samwaseda (commented):
Sorry I'm still struggling with other work to do but I try to contribute a bit more little by little.

pyiron_contrib/workflow/io.py
liamhuber: `__getattr__` on this class just looks in the dictionary of owned channels, so this is disconnecting the channel named `run` if it exists. This is _useful_ because every `Node` adds a `signals.input.run` and `signals.output.ran` to itself at instantiation.

In the long run it might be nice to do something cute [like Marvin did here](https://github.com/liamhuber/atomistics/blob/1a6dbfe96600b82cf7db4e4438cc995065a543d8/atomistics/calculators/interface.py#L15) and have the "immortal channels" rigorously defined somewhere. In the meantime, while I admit that the usefulness is opaque without knowledge of other classes, I'm comfortable with this because the behaviour is so explicit -- `disconnect_run` attempts to disconnect run 😂

pyiron_contrib/workflow/channels.py
liamhuber: Ooops, I forgot to switch branches and added this to the next PR in the stack. Good idea and done!

liamhuber:
> Sorry I'm still struggling with other work to do but I try to contribute a bit more little by little.

No worries! I'm very much looking forward to getting more regular feedback and contributions from you again, but I get that there's other stuff on your plate.

I will probably merge this stack imminently so that it's all on `HEAD` for the meeting on Monday.

pyiron_contrib/workflow/node.py
```diff
@@ -277,9 +277,18 @@ def server(self, server: Server | None):
...
+
+        Returns:
+            [list[tuple[Channel, Channel]]]: A list of the pairs of channels that no
+                longer participate in a connection.
+        """
+        destroyed_connections = []
+        destroyed_connections.extend(self.inputs.disconnect())
+        destroyed_connections.extend(self.outputs.disconnect())
+        destroyed_connections.extend(self.signals.disconnect())
+        return destroyed_connections
```
samwaseda, line 291:
```suggestion
        return self.inputs.disconnect() + self.outputs.disconnect() + self.signals.disconnect()
```

If the line is not too long

Issue #836, opened 2023-09-05
"pull"-mode


liamhuber:
@JNmpi sold me a while ago on the merits of using a "pull" paradigm instead of a "push" paradigm, but I couldn't see a way to get the technicalities sorted out. I think I see a path through now and will be working on this topic this week.

Most of the classes look the same, and I think we do still need signals for making cyclic graphs work (and indeed, switches in general), but I think this will greatly streamline things and make "easy things easy" while not making hard things any harder than they already are.

Here's my thoughts so far on the class behaviour:

- `OutputData.pull()`
  - Calls `self.parent.outputs.pull(): Outputs.pull()`
  - `Outputs.pull()` manually sets `OutputData.value` for each child
  - Return `self.value` once the `Outputs.pull()` call resolves
- `Outputs.pull()`
  - Calls `self.node.pull(): Node.pull()`
  - Uses the returned dictionary to update the values of all channels
- `Node.pull()`
  - Calls `self.inputs.pull(): Inputs.pull()`
  - Hashes the resulting dictionary of input data 
  - Compares the hash to the hash for the last call...
    - If there is no such hash, or hashes differ, run with input, returns result of run function
    - Else return existing output `self.outputs.to_dict()`
    - This prevents unnecessary running when input hasn't changed
    - Need a flag like `Node.always_rerun` in case we _want_ to run again on the same input! E.g. random number generation
- `Inputs.pull()`
  - Calls `.pull()` on each `InputData` child
  - Returns children's values as a dictionary for use in the node
- `InputData.pull()`
  - For each connection in `self.connections: list[OutputData]` call `.pull()`
    - The first one that returns something other than `NotData` gets used
    - _Somewhere_ we need to do more clever grouping, so that we directly call `Outputs.pull()` so that groups of output can get updated at once without multiple graph traversals where we discover that the input hasn't changed...
    - This is a change that makes `connections` and _ordered list_ so we will always first try to draw data from the 0th connection, then use the 1st as a fall-back, etc.
  - If `len(self.connections) == 0` return `self.value`
    - If no default exists, and no value was set manually, e.g. by the user of the workflow, the initial `NotData` value for `DataChannel.value` gets returned -- obviously this will break something somewhere, either by raising an error or by keeping some form of the `Channel/IO/Node.ready` attribute that will stop execution.

[...]

PR #840, opened 2023-09-09
Simplest dag automation

liamhuber:
In the case of a DAG, execution flow can be sufficiently determined from the topology of the data flow graph. For cyclic graphs I still haven't figured out a way to avoid `run`/`ran` execution signals, and anyhow I still kind of like the way they keep the graph information embedded at the node level without giving the workflow some kind of special supremacy. Thus, I want to use DAG data flow analysis to automate the construction of run signals.

This is a _suuuuper_ rough implementation of that using the lightweight `toposort` library. It creates a serial execution flow that guarantees that all a node's dependencies have executed (and pushed their data forwards) before the node runs. Users can still (and in the case of cyclic graphs _need to_) fall back to manually specifying the execution signals with the `automate_execution` attribute/init argument.

It's rough/early in that I haven't done tests etc., but I want to show it working ASAP. After getting this polished and working as desired, the final step is more sophisticated exploitation of the data flow topology to efficiently run multiple nodes in parallel (e.g. when they have an executor) instead of the serial flow here. That might require switching over to the more powerful `networkx` library, but I wanted to keep dependencies as lightweight as possible at this stage. This is the most primitive resolution of the final TODO item over in #836.

The following DAG example now works without users ever having to worry about signals:

```python

@Workflow.wrap_as.single_value_node("out")
def foo(x, y):
    return x + y

wf = Workflow("dag")
wf.n1l = foo(1, 2)
wf.n1r = foo(3, 4)
wf.n2l = foo(10, wf.n1l)
wf.n2m = foo(wf.n1l, wf.n1r)
wf.n2r = foo(wf.n1r, 20)
wf.n3l = foo(100, wf.n2l)
wf.n3r = foo(wf.n2r, 200)
wf.n4 = foo(wf.n2m, wf.n3r)

wf.run()
>>> {'n3l__out': 113, 'n4__out': 237}
```

Cyclic graphs throw an error:

```python
wf = Workflow("cyclic")

wf.n1 = foo()
wf.n2 = foo()
wf.n1.inputs.x = wf.n2
wf.n2.inputs.x = wf.n1

wf.run()
>>> CircularDependencyError
>>> ...
>>> RuntimeError: Detected a cycle in the data flow topology, unable to automate the execution of non-DAGs.
```

TODO:
- [x] Tests for everything
- [x] Differentiate workflows (dynamically create execution order at run call) and macros (create execution order once at instantiation)
- [x] Guarantee that nesting works as intended
- [x] Refactor for maintainability
- [x] Improve UX, e.g. with more informative exception messages
- [x] Update docstrings and demo notebook

TODON'T (next PR(s)):
- A more sophisticated treatment of dependencies so that parallel nodes run efficiently
- Nicely handle running connected nodes outside a workflow context, probably with the introduction of `pull` and `push` methods so that `run` still just runs with the existing input


liamhuber:
Ok, I'm content with this now. It's not efficient when nodes can run on parallel processes, and the API for while loop calls is butt-ugly, but it's all working as desired now. In particular
- Default behaviour is that data DAGs will automatically generate their execution flow (via starting nodes and run signals)
- Acyclic macros where execution flow needs to be defined can be safely and robustly nested inside other workflows/macros
- Composites (workflows and macros) can now _remove_ children's unconnected IO from composite-level IO (in addition to the existing power to add children's connected IO to composite IO)

PR #885, opened 2023-10-09
TrainingContainer: Allow to customize HDF keys

pmrv:
Allows to change in the location from hdf file where training data is read from by TrainingContainer.include_job.

This also adds a Vasp specific work around to read the smeared energy from the file instead of the generic energy_pot (which is the sigma to zero extrapolated energy), which previously resulted in inconsistent energies and forces.

### DIFF
--- a/pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
+++ b/pyiron_contrib/atomistics/atomistics/job/trainingcontainer.py
@@ -373,6 +373,29 @@ class TrainingPlots(StructurePlots):
         plt.xlabel(r"Force [eV/$\mathrm{\AA}$]")
 
 
+_HDF_KEYS = {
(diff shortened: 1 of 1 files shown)


pmrv:
@Leimeroth Pinging you, because I don't how much you use the `TrainingContainer` but the bug with the VASP energies might have affected your calculations as well.

Leimeroth:
Not using the TrainingContainer much right now, but still the different vasp energies cause us some headache. We didn't see much difference for sigma values converged to the vasp recommendation of <1meV/atom though.

pmrv:
Great, as long as you are aware, that's ok.

jan-janssen:
@pmrv Can you transfer this pull request to https://github.com/pyiron/pyiron_potentialfit ?

PR #886, opened 2023-10-10
Include pyiron-data in the dependencies

jan-janssen:


liamhuber:
@jan-janssen just FYI, with repos using the centralized CI you only need to update the dependency in `.ci_support/environement.yml` and a bot will take care of docs and binder. Of course no harm is caused doing it manually either, I'm just trying to sell `pyiron/actions` a little 😉

liamhuber:
Indeed, tests run fine with `pyiron-data` explicitly included.

jan-janssen:
@liamhuber I do not understand why you closed this pull request, as it fixes the current issue until we have a better solution.

jan-janssen:
Only include `pyiron-data` in the notebooks environment https://github.com/pyiron/pyiron_contrib/pull/893

PR #887, opened 2023-10-16
Move potential fitting to pyiron_potentialfit

jan-janssen:


jan-janssen:
Waiting for https://github.com/pyiron/pyiron_contrib/pull/886 to be merged first.

jan-janssen:
@pmrv I added the import statements for the job classes listed in `pyiron_contrib/__init__.py`. Do you need any other classes?

pmrv:
Everything that has to/from hdf methods would need to be available at least for a certain amount of time.  In particular `TrainingStorage` and subclasses from it and `FlattenedStorage` that are defined for the atomicrex and runnner codes.  I will try to have a look later to see if there's anything else.

@Leimeroth Anything you can think of?

pmrv:
@Leimeroth We are working on a fix over in [base](https://github.com/pyiron/pyiron_base/pull/1234) that will fix the importing of old jobs and update the HDF5 storage.  Once this is in (hopefully this week), I would like to merge this.  Any other concerns you might have here or local changes to atomicrex that would need to be merged before we can make the split?

Leimeroth:
> Everything that has to/from hdf methods would need to be available at least for a certain amount of time. In particular `TrainingStorage` and subclasses from it and `FlattenedStorage` that are defined for the atomicrex and runnner codes. I will try to have a look later to see if there's anything else.
> 
> @Leimeroth Anything you can think of?

Will the fix in base resolve this? If not I guess basically everything breaks since everything defines to_hdf().

pmrv:
> > Everything that has to/from hdf methods would need to be available at least for a certain amount of time. In particular `TrainingStorage` and subclasses from it and `FlattenedStorage` that are defined for the atomicrex and runnner codes. I will try to have a look later to see if there's anything else.
> > @Leimeroth Anything you can think of?
> 
> Will the fix in base resolve this? If not I guess basically everything breaks since everything defines to_hdf().

Yes, the fix will define a [function](https://github.com/pyiron/pyiron_base/pull/1234/files#diff-c1466f4a0a1bce7c0f89cead5cc8987a964b532ce928b3b323a1092b9c17b410R51) that we can use from pyiron_contrib __init__ so that objects written from pyiron_contrib will be redirected to the correct new modules in `potentialfit`.

PR #896, opened 2023-10-24
Add generic job factories

pmrv:
A JobFactory is a light weight object that represents a certain type of pyiron job.  It can be used by abstract workflows to run jobs without knowing what the reference model and the detailed parameters e.g. for the HPC queue are, without having to pass all of this information through all layers of the workflow.

### DIFF
--- a/pyiron_contrib/jobfactories/__init__.py
+++ b/pyiron_contrib/jobfactories/__init__.py
@@ -136,11 +136,6 @@ class JobFactory(HasStorage, ABC):
         copy.storage.update(self.storage.copy())
         copy.project = self.project
         return copy
-        # copy = self.__class__()
-        # for k in self.storage:
-        #     copy.storage[k] = self.storage[k]
-        # copy.project = self.project
-        # return copy
 
     def set_input(self, **kwargs):
         """
--- a/pyiron_contrib/jobfactories/__init__.py
+++ b/pyiron_contrib/jobfactories/__init__.py
@@ -46,7 +46,7 @@ class JobFactory(HasStorage, ABC):
     >>> strains = np.linspace(-0.5, 1.5, 500)
 
     >>> for eps in strains:
-    ...   j.run(f'my_lammps_{eps}', modify=lambda: None, structure=cu)
+    ...   j.run(f'my_lammps_{eps}', modify=lambda j: j['user/strain'] := eps and j, structure=cu.apply_strain(eps-1, return_box=True))
 
     is much easier to understand than a loop around the above "standard"
     construct.  While one could use functions to save a similar amount of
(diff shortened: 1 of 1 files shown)


pyiron_contrib/jobfactories/__init__.py
```diff
@@ -0,0 +1,364 @@
...
+    ...   j = None
+
+    For a single job this is ok syntax sugar, but for large numbers of
+    submitted jobs this can become a substantial simplification, i.e.
+
+    >>> import numpy as np
+    >>> strains = np.linspace(-0.5, 1.5, 500)
+
+    >>> for eps in strains:
+    ...   j.run(f'my_lammps_{eps}', modify=lambda: None, structure=cu)
```
pmrv, line 49:
```suggestion
    ...   j.run(f'my_lammps_{eps}', modify=lambda j: j['user/strain'] := eps and j, structure=cu.apply_strain(eps-1, return_box=True))
```

pmrv:
This may be a bit cryptic (the example in the docstring might help or it might not), but I use these objects in a lot of my work now and I need them to live somewhere.  If this is not wanted in contrib, I can put them up separately.

pyiron_contrib/jobfactories/__init__.py
```diff
@@ -0,0 +1,364 @@
+from pyiron_base import HasHDF, HasStorage, GenericJob
```
pmrv, line 1:
```suggestion
from pyiron_base import HasStorage, GenericJob
```

pyiron_contrib/jobfactories/__init__.py
```diff
@@ -0,0 +1,364 @@
...
+
+    @property
+    def attr(self):
+        """
+        Access to attributes, that should be set on the job after creation.
+        """
+        return self.storage.create_group('attributes')
+
+    @property
+    def project(self):
```
pmrv, line 85:
```suggestion
    def project(self) -> Project:
```

pyiron_contrib/jobfactories/__init__.py
```diff
@@ -0,0 +1,364 @@
...
+        """
+        Project that jobs should be created in by :meth:`.make`.
+        """
+        return self._project
+
+    @project.setter
+    def project(self, value):
```
pmrv, line 92:
```suggestion
    def project(self, value: Project):
```

pyiron_contrib/jobfactories/__init__.py
```diff
@@ -0,0 +1,364 @@
...
+        copy = type(self)()
+        copy.storage.clear()
+        copy.storage.update(self.storage.copy())
+        copy.project = self.project
+        return copy
+        # copy = self.__class__()
+        # for k in self.storage:
+        #     copy.storage[k] = self.storage[k]
+        # copy.project = self.project
+        # return copy
```
pmrv, line 143:
```suggestion
```

PR #905, opened 2023-11-13
add tight layout

Leimeroth:
Prevent overlap of title and axis description in eam plot


Leimeroth:
Since I only added the fig.tight_layout() line I guess the reason for failing tests are the ModuleNotFoundErrors. Prrobably should be fixed in pyiron_atomistics or atomistics. @jan-janssen could you have look where it should be added as dependency?

PR #926, opened 2023-12-01
[new job type] MD analysis tool for electrochemistry repo

ahmedabdelkawy:


pyiron_contrib/__init__.py
```diff
@@ -58,6 +58,7 @@
         "PigletMD": "pyiron_contrib.atomistics.ipi.ipi_jobs",
         "FitsnapJob": "pyiron_contrib.atomistics.fitsnap.job",
         "QuasiHarmonicApproximation": "pyiron_contrib.atomistics.atomistics.master.qha",
+        "MDAnalysis": "pyiron_electrochemistry.md_analysis"
```
jan-janssen, line 61: You can import `from pyiron_base import JOB_CLASS_DICT` in `pyiron_electrochemistry` and then just load the module there: 
```
JOB_CLASS_DICT.update({"MDAnalysis": "pyiron_electrochemistry.md_analysis"})
```

ahmedabdelkawy: Okay that worked. It was added in the __init_ file and not in the md_analysis class.

PR #948, opened 2024-01-05
Tinystore Draft

pmrv:
Includes #936 halfway.

@liamhuber There's still some problems with [h5io_browser](https://github.com/h5io/h5io_browser/releases) that I use to access the files.  I think you will need to force python=3.10 to avoid a bug there that otherwise breaks reading data,


pmrv:
Instead of `ProjectAdapter` you'll want to use the `FilesystemProject` or `SingleHdfProject` classes.

liamhuber:
Thanks!

> @liamhuber There's still some problems with [h5io_browser](https://github.com/h5io/h5io_browser/releases) that I use to access the files. I think you will need to force python=3.10 to avoid a bug there that otherwise breaks reading data,

No worries, [`pyiron_workflow` already specifies python>=3.10 anyhow](https://github.com/pyiron/pyiron_workflow/blob/1ea4ff69e0eed921e625cbef519e93d72eec0363/setup.py#L24) (for type-hinting syntax IIRC).

pmrv:
Ah, but in this case 3.11 is the one that causes the problem.  I will patch it by Monday, I think.

liamhuber:
> Ah, but in this case 3.11 is the one that causes the problem. I will patch it by Monday, I think.

Indeed, that is more of a problem 😂 If you can patch that would be stellar, and Monday is fine -- despite my grand ambitions to work on this I spent most of my morning [trying to track down](https://github.com/pyiron/pyiron_workflow/pull/149) a crash in the CI that I couldn't reproduce on my local machine 🤦‍♂️

PR #949, opened 2024-01-08
Tinybase: New Input/Output, Renamed Executors, Storage/Directory brokering by Project

pmrv:
Supersedes #948 and #936.

- Input and output classes are based on dataclasses now; this also changed how the Output is generated by tasks.  Previously they got an output instance injected and were expected to populated it, now they are expected to return it.  This simplified some of the "workflow" nodes like the `SeriesTask` and makes it e.g. possible to run Murnaghan calculations with static, minimizing and molecular dynamics quite generically.
- Executors are called `Submitter`s now and work with anything that follows the `concurrent.futures` interface.  In the future I'll likely completely remove them.
- I added a new storage interface based on h5io directly and added new project implementations that e.g. store everything in a single file.
- Tasks are now aware of their working directory and ask for one from a Project.  This allows the project to fully control filesystem access of its jobs.


pmrv:
@liamhuber The details of the bug I mentioned are described [here](https://github.com/h5io/h5io/pull/69).  Since in the snippet we discussed you used __getstate__/__setstate__ on your own outside of h5io, it might work for you anyway though.

pmrv:
The error now is related to pip/setuptools, googling around this can be caused when some dependencies are not compatible to the [latest](https://github.com/readthedocs/readthedocs.org/issues/10286) version of setuptools.  In fact downgrading it seems to resolve the issue locally for me, but no idea what "not compatible" actually means in this context and how to fix it for the CI.

@liamhuber I should hope that if you just check this out locally and install the dependencies with conda/mamba it should work. :|

.binder/environment.yml
```diff
@@ -20,6 +20,7 @@ dependencies:
 - aws-sam-translator =1.77.0
 - pympipool =0.7.9
 - distributed =2023.12.1
+- h5io_browser =0.0.4
```
jan-janssen, line 23: Is there a specific reason to use `0.0.4` rather than `0.0.5`?

jan-janssen: In addition, you could also try to update the numpy version to `numpy =1.26.3` as that is what is supported by `h5io_browser`, `pyiron_base` and so on. Maybe those lead to very strange resolution of the dependencies.

pyiron_contrib/tinybase/project/h5io.py
```diff
@@ -0,0 +1,84 @@
...
+    Behaves likes a :class:`~.FilesystemProject` but stores all jobs in a single HDF5 file.
+    """
+
+    def _get_hdf(self):
+        return H5ioStorage.from_file(self, os.path.join(self.path, "project.h5"))
+
+    def create_storage(self, name):
+        if not os.path.exists(self.path):
+            os.makedirs(self.path, exist_ok=True)
+        return self._get_hdf()[name]
```
liamhuber, line 78:
```suggestion
    def _get_hdf(self, filename):
        return H5ioStorage.from_file(self, os.path.join(self.path, filename))

    def create_storage(self, name, filename="project.h5"):
        if not os.path.exists(self.path):
            os.makedirs(self.path, exist_ok=True)
        return self._get_hdf(filename)[name]
```

Experimenting with this in `pyiron_workflow` it feels funny that the h5 file is named "project.h5"; are you comfortable making it a variable?

.binder/environment.yml
pmrv, line 23:
```suggestion
- h5io_browser =0.0.5
```

pmrv: No, I think that's just what was there when I made the change locally.

pyiron_contrib/tinybase/project/h5io.py
pmrv: I wouldn't want it in the `create_storage` method, because imo the file should be constant after project instantiation, but it could go through `__init__` without problem.

I realize thinking about the response that the concepts and division of responsibilities for the interfaces here is nowhere written down so far.  I'll try to make a little chart during the retreat or so.

setup.py
```diff
@@ -54,9 +54,9 @@
         ],
         'tinybase': [
             'distributed==2023.12.1',
-            'pympipool==0.7.9'
+            'pympipool==0.7.9',
```
jan-janssen, line 57: @pmrv The `setuptools` issue was a missing `,` after `pympipool`. Then the two strings `'pympipool==0.7.9'` and `'h5io_browser==0.0.4',` are combined to once string `'pympipool==0.7.9h5io_browser==0.0.4'` which does not follow the requirements for the `setup.py` file. 
```
error in pyiron_contrib setup command: 'extras_require' must be a dictionary whose values are strings or lists of strings containing valid project/version requirement specifiers.
```

pmrv: Ah, yikes.  Thanks for catching that!

jan-janssen:
```
ValueError: Failed to restore object with: Failed to restore object with: module 'builtins' has no attribute 'method'
```

pmrv:
> ```
> ValueError: Failed to restore object with: Failed to restore object with: module 'builtins' has no attribute 'method'
> ```

That's error message that I got from h5io prior to https://github.com/h5io/h5io/pull/69, I'll bump deps.

.binder/environment.yml
```diff
@@ -20,6 +20,7 @@ dependencies:
 - aws-sam-translator =1.77.0
 - pympipool =0.7.9
 - distributed =2023.12.1
+- h5io_browser =0.0.5
```
pmrv, line 23:
```suggestion
- h5io_browser =0.0.6
```

.ci_support/environment.yml
```diff
@@ -20,3 +20,4 @@ dependencies:
 - aws-sam-translator =1.77.0
 - pympipool =0.7.9
 - distributed =2023.12.1
+- h5io_browser =0.0.5
```
pmrv, line 23:
```suggestion
- h5io_browser =0.0.6
```

docs/environment.yml
```diff
@@ -22,3 +22,4 @@ dependencies:
 - aws-sam-translator =1.77.0
 - pympipool =0.7.9
 - distributed =2023.12.1
+- h5io_browser =0.0.5
```
pmrv, line 25:
```suggestion
- h5io_browser =0.0.6
```

setup.py
```diff
@@ -54,9 +54,9 @@
...
+            'h5io_browser==0.0.4',
```
pmrv, line 58:
```suggestion
            'h5io_browser==0.0.6',
```

liamhuber:
Test failure is just `pyiron_base` demanding a lower `h5io_browser`.

```
├─ h5io_browser 0.0.6**  is requested and can be installed;
...
└─ pyiron_base 0.6.20**  is installable and it requires
   └─ h5io_browser >=0.0.2,<=0.0.5  with the potential options
```

liamhuber:
Question: Why are the windows tests so slow?

Windows takes 12 minutes and counting, while other OSs take <5 minutes. I compared the windows log to a couple ubuntu logs, and the actual tests themselves (55 OK + 1 skipped in all cases) _do_ take longer, but we're talking 8s instead of 5s -- insignificant.

I didn't see where the actual time is being eaten, but it must be somewhere in the env setup since the tests themselves are OK.

Ultimately, I find this acceptable. The windows tests are annoyingly slow, but it's something like 3x not 10x, and the windows _operation_ times are not horrible (at least in what is unit tested). It's a bit puzzling, but I can live with it.

liamhuber:
Omitting the ASE notebook over on #963 did indeed allow the rest of the notebook tests to complete ok, and quickly.

The Basic, Lammps, and Shell notebooks all fail exactly as shown above. Most of these look to me like they're just hiccups in either the code or example calls, and that providing missing data, arguments, etc. will clean things up.

Interestingly, the TinyBase notebook ran totally fine on the CI. I don't have any knee-jerk response to the error message I got and am not sure why this is. It is also not high-priority for me to figure it out.

Regarding the ASE notebook, the pickling error above smells a lot like pickle's difficulties with serializing stuff defined inside a jupyter notebook.

@pmrv, I don't plan to patch any of these problems imminently, but if I do get to it then it will be in a stacked PR and I'll open it as a draft early in the process so we don't duplicate any work. In the meantime I don't want to migrate the submodule from here to the workflows while the demos are broken. 
  Immediately, I will just proceed on the current path of delaying the contrib import until save-time, and may spoof `Storable` as we discussed. In the near-term, we have the option of migrating just the bits that are needed _right now_ (i.e. the storage module) and migrating the rest (database interface abstractions, etc.) later as needed. I have mixed feelings about this as (a) I want the stuff, but (b) breaking it apart has a good chance of causing more headaches than it solves. Opinion?

pmrv:
Thanks for the detailed check!  Sounds like I was careless and only checked the working directory changes on TinyJob.ipynb.  Will update next week when we're back.

liamhuber:
```
...
Input Notebook:  notebooks/tinybase/Shell.ipynb
...
Exception encountered at "In [5]":
AttributeError                            Traceback (most recent call last)
Cell In[5], line 1
----> 1 out.stdout
...
```

liamhuber:
> ...
> Input Notebook:  notebooks/tinybase/Shell.ipynb
> ...
> ---------------------------------------------------------------------------
> Exception encountered at "In [5]":
> ---------------------------------------------------------------------------
> AttributeError                            Traceback (most recent call last)
> Cell In[5], line 1
> ----> 1 out.stdout
> ...

This was because the shell task now requires the `working_directory` input, so the task never ran because its input was incomplete

PR #962, opened 2024-01-15
Lpatch tinydata

liamhuber:
Patching test failures on #949

### DIFF
--- a/notebooks/tinybase/ASE.ipynb
+++ b/notebooks/tinybase/ASE.ipynb
@@ -151,16 +151,6 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 12,
-   "id": "f1bfdf31-84ee-42bd-a230-d990f2cc6116",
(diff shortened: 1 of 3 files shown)


liamhuber:
Notebook tests are just hanging (>30mins).

In `pyiron_workflow`, `pympipool` requires a modification of the path so that the unit tests can be found... perhaps something similar is happening here with the notebooks? @jan-janssen, as far as I can see the `pympipool` tests never run the example notebook there, so maybe you have some insight on this.

For now I'm just going try-and-see by getting rid of `pympipool` from the notebook.

liamhuber:
Windows and notebooks runs are both still going after 12 minutes 👎

liamhuber:
Windows took a while (20 mins) but finished. Notebooks got through checkout but were "pending" on the actual run for over an hour

pmrv (approved):
lgtm

PR #968, opened 2024-01-23
Lighten the import of pyiron_contrib

niklassiemer:
In the main init, only pyiron_base is imported to update the JOB_CLASS_DICT.


liamhuber (approved):
Thanks, Niklas! Code looks a-ok. I checked out the branch and this is bringing the walltime for `import pyiron_contrib` from ~20s down to ~2s on my machine 🚀 🚀

PR #978, opened 2024-02-02
Update shell notebook

liamhuber:
The `ShellTask` now requires working directory input, and I un-hardcoded the resource path.

### DIFF
--- a/notebooks/tinybase/Shell.ipynb
+++ b/notebooks/tinybase/Shell.ipynb
@@ -11,7 +11,7 @@
     {
      "data": {
       "application/vnd.jupyter.widget-view+json": {
-       "model_id": "d4e54a8ed92444e48502ea9c35497c00",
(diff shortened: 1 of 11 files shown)


liamhuber:
Shell notebook again

```
Exception encountered at "In [47]":
AttributeError                            Traceback (most recent call last)
Cell In[47], line 1
----> 1 out.returncode

AttributeError: 'NoneType' object has no attribute 'returncode'
```

liamhuber:
Cell 47 is the first output access after we try to access the local resources, so I'm guessing my relative path hack of `state.settings.resource_paths.insert(0, str(cwd / "resources"))` is failing because the `cwd` is differing from the actual notebook location on the CI

liamhuber:
Indeed, I can reproduce the error in that cell when the resource path has not been added

PR #996, opened 2024-02-16
Bump base, atomistics, and h5io_browser

liamhuber:
Accidentally committed this directly to main without noticing and it passed all the tests there, so I guess it should be good 🤦‍♂️ 

Now as a branch...

### DIFF
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -1,27 +1,4 @@
 channels:
 - conda-forge
 dependencies:
-- ase =3.22.1
-- coveralls
-- coverage
-- codacy-coverage
-- matplotlib =3.8.2
-- numpy =1.26.4
-- pyiron_base =0.7.8
-- pyiron_atomistics =0.4.15
-- pyparsing =3.1.1
-- scipy =1.12.0
-- seaborn =0.13.2
-- scikit-image =0.22.0
-- randspg =0.0.1
-- boto3 =1.34.39
-- moto =5.0.1
-- pycp2k =0.2.2
-- aws-sam-translator =1.77.0
-- pympipool =0.7.13
-- distributed =2024.2.0
-- h5io_browser =0.0.9
-- python >= 3.10
-- lammps
-- nglview >=3.0.8
-- pyiron-data >=0.0.25
+- pip: []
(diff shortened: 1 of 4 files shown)


liamhuber:
> > Accidentally committed this directly to main without noticing and it passed all the tests there, so I guess it should be good 🤦‍♂️ 
> > 
> > Now as a branch...
> 
> Maybe we should finally protect the main branch from direct push ;) 

Yeah, gonna do that as soon as I get home! I understand the rationale for the rule now 😂

liamhuber:
> > Accidentally committed this directly to main without noticing and it passed all the tests there, so I guess it should be good 🤦‍♂️
> > Now as a branch...
> 
> Maybe we should finally protect the main branch from direct push ;)

@niklassiemer did you add the rule in the last couple hours? My reading of the currently-in-place rule is that it should have stopped what I did, so if you just added it then 🚀 but if it's been there all along then I've misunderstood the rules and we need something stricter!

niklassiemer:
> > > Accidentally committed this directly to main without noticing and it passed all the tests there, so I guess it should be good 🤦‍♂️
> > > Now as a branch...
> > 
> > Maybe we should finally protect the main branch from direct push ;)
> 
> @niklassiemer did you add the rule in the last couple hours? My reading of the currently-in-place rule is that it should have stopped what I did, so if you just added it then 🚀 but if it's been there all along then I've misunderstood the rules and we need something stricter!

No, I did not alter the rules here! We might need to explicitly evaluate what is possible on a dummy repo...

liamhuber:
> No, I did not alter the rules here! We might need to explicitly evaluate what is possible on a dummy repo...

Ah, yikes. I just went ahead and used this repo, but modifying just a comment in some random file -- I'm not super fussed about keeping the git history pristine.

Turns out we need to _also_ [check a box](https://stackoverflow.com/questions/58541814/how-to-protect-branch-with-require-pull-request-reviews-before-merging/58542669#58542669) preventing admins from ignoring the rules. Because otherwise they ignore them by default 😝

niklassiemer:
:lol: fair enough! Just check the box every time you stumble on it then - I think we are just as prone to accidental pushes as everyone else xD

liamhuber:
Yeah 😂 I didn't go through and click it exhaustively, but I agree we should click it as it comes up and I did for a couple of the usual suspects

PR #1002, opened 2024-02-26
Add the HandyMan

pmrv:
Adds some classes that can automatically repair crashed pyiron jobs.  

```python
# project with crashed jobs
pr = Project(...)

from pyiron_contrib.repair import HandyMan
hm = HandyMan()
# goes over all jobs in pr and finds those that it knows how to repair, then restarts them
cs = hm.fix_project(pr)
cs.fixing # dictionary of job ids of restarted jobs and by what tools
cs.failed # list of jobs that were attempted to restarted but threw an error
cs.hopeless # list of jobs that we don't know how to fix
```

Included tools:
1. jobs that ran into timeout
2. master jobs that had crashed children, but which were meanwhile repaired
3. very long list of VASP specific fixes and work arounds.

### DIFF
--- a/pyiron_contrib/repair/__init__.py
+++ b/pyiron_contrib/repair/__init__.py
@@ -451,8 +451,8 @@ class VaspZpotrfZtrtri(VaspTool):
 
     def match(self, job):
         return super().match(job) and match_in_error_log(PartialLine(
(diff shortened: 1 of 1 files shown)


pmrv:
Maybe add something that adds the docstring so users know what each tool does.

PR #1009, opened 2024-02-29
splitting up job entries into different tables

tnecnivkcots:
As long time discussed and awaited, the splitting of job entries into different tables can now be seen and tested with the branch [splitted_db_tables](https://github.com/pyiron/pyiron_contrib/tree/splitted_db_tables).

What is not already working:
- master and parent relationships, as they point to the same primary key in the jobs table and lead to an error
- using just one session, for the complete process of adding a job

There are also some additional questions:
- do we want to have a view mode, as in pyiron_base?
- do we want to check for the length of the chemicalformula?

Some documentation diagrams will follow.

### DIFF
--- a/pyiron_contrib/tinybase/database.py
+++ b/pyiron_contrib/tinybase/database.py
@@ -3,58 +3,426 @@ from collections import namedtuple
 import os.path
 from typing import List
 from typing import Optional
-from sqlalchemy import ForeignKey, String, Integer, Column, create_engine
+from sqlalchemy import \
+    create_engine, \
+    Table, Column, \
+    ForeignKey, UniqueConstraint, \
+    Integer, BigInteger, String, TIMESTAMP, \
+    select
 from sqlalchemy.exc import MultipleResultsFound, NoResultFound
 from sqlalchemy.orm import declarative_base, relationship
 from sqlalchemy.orm import Session
 from sqlalchemy.pool import StaticPool
 
+import json
+
 import pandas as pd
 
+
+def get_sqlalchemy_class_doc(tablename, classname, **kwargs):
+    """
+    produces the docstrings from the inserted arguments for sqlalchemy classes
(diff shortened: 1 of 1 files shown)


jan-janssen:
Hi @tnecnivkcots , do you think it makes sense to present this in next weeks pyiron meeting ? If so can you add a short comment to https://github.com/orgs/pyiron/discussions/218 . In addition, it would be great to have some examples and or unit tests to see how the new interface compares to the previous interface. Is the new interface backwards compatible? How can a user switch their local environment to use the new database interface?

pmrv:
So currently one cannot switch from the old to the new interface.  My plan is to stabilize the new interface here first a little bit and then make DatabaseAccess from base conform to it.  Switching over the base code would then be simple, but would also need some compatibility setup from our database, either by defining some table views or just migrating the whole database on a maintenance weekend.

Notebooks are failing right now because the rest of tinybase is not adapted yet, but once I've done that we can play a bit with the setup in a tinybase only setup.

tnecnivkcots:
In docs folder is now a subdirectory [uml_diagrams](https://github.com/pyiron/pyiron_contrib/tree/splitted_db_tables/docs/uml_diagrams), for concept documentation (in .md and .png-image). For the first there is just a database schema.

[database schema](https://github.com/pyiron/pyiron_contrib/tree/splitted_db_tables/docs/uml_diagrams/db_schema.png)

tnecnivkcots:
> Hi @tnecnivkcots , do you think it makes sense to present this in next weeks pyiron meeting ? If so can you add a short comment to https://github.com/orgs/pyiron/discussions/218 . In addition, it would be great to have some examples and or unit tests to see how the new interface compares to the previous interface. Is the new interface backwards compatible? How can a user switch their local environment to use the new database interface?

I think it will be a bit of a work in progress for the presentation. In particular, using it on our cmti resources will require the adjustment of our database setup.

PR #1073, opened 2024-06-03
[patch] replace logger with snippets version

liamhuber:
i.e. pyiron_base.state.logger.logger with pyiron_snippets.logger.logger. The original location leaves a re-direct for backwards compatibility.

PR #1075, opened 2024-06-04
Update import paths in elastic test

pmrv:


pmrv:
Notebook failure is the same as in #1073 and fixed there anyway.

PR #1116, opened 2024-07-31
Fix CI problems

samwaseda:


samwaseda:
I don’t understand why it’s happening:

```
AttributeError: 'ShellOutput' object has no attribute 'energy_pot'
```

It’s coming from [`Lammps.ipynb`](https://github.com/pyiron/pyiron_contrib/blob/main/notebooks/tinybase/Lammps.ipynb), but I cannot reproduce it in my environment. Do you have an idea? @pmrv

samwaseda:
So in the end instead of solving the problem apparently I made more problems. Now, we got also the problem with coveralls and pycp2k which pip says is not available on the system. I’ll try to look into it tomorrow since it’s late now…

pmrv:
#1119 should take care of this, at least the CI runs again.

PR #1119, opened 2024-08-02
Bump base/atomistics/pymatgen together

pmrv:


samwaseda:
I don't really understand how this solved the problem? Is there a clear explanation?

pmrv:
For the `pycp2k` problem, we just dropped the package from the env, because we figured it's an issue with their pip package maybe not supporting python3.12 yet.
The notebook failure was caused by there not being a `lmp` executable even when lammps was installed, @jan-janssen said this may have been an old lammps package that was pulled by our env, since we don't pin the version there.
The other changes are just updating the action to the latest versions, I'm not sure if that did something indirectly.

samwaseda:
God I wasted 2 days trying to figure out the problem. This shouldn't happen

PR #1126, opened 2024-08-08
[patch] Relocate `pyiron_workflow.job`

liamhuber:
Move the module here [using git copy](https://gist.github.com/liamhuber/73a5ff0592d250686245741283c52702) and update dependencies. Note that this is waiting on a release of `pyiron_workflow` that doesn't exist yet -- depending on existing releases would make a cyclic dependency.


liamhuber:
`pyiron_workflow` and `pyiron_base` conflict over `pyiron_snippets`:

```
  ├─ pyiron_base 0.9.11**  is installable and it requires
  │  └─ pyiron_snippets 0.1.3 , which can be installed;
  └─ pyiron_workflow 0.10.0**  is not installable because there are no viable options
     ├─ pyiron_workflow 0.10.0 would require
     │  └─ pyiron_snippets 0.1.4 , which conflicts with any installable versions previously reported;
     └─ pyiron_workflow 0.10.0 would require
        └─ pyiron_base 0.9.12 , which conflicts with any installable versions previously reported.
```

I guess I'll go try bumping base

